PR #27, opened 2017-09-20
Task: Simplify development

kenjones-cisco:
The base directory has all shell scripts in scripts/
and all example/test files in examples/.
A Makefile provides all the commands with helpful information.
If a developer simply types `make` then vendor is properly updated,
the code is formatted, linted, tested, built, acceptance test run,
and installed.

Linting errors resolved.
Ignored test case (`TestParsePath`) updated to work as expected.


mikefarah:
this is great, thank you!

Issue #17, opened 2017-07-26
push to exist array structure


davidkhala:
In my case, I have to write yaml arrays to file in this way

```
    yaml w -i $COMPOSE_FILE services.$peerContainer.environment[0] a
    yaml w -i $COMPOSE_FILE services.$peerContainer.environment[1] b  
    yaml w -i $COMPOSE_FILE services.$peerContainer.environment[2] c  
    yaml w -i $COMPOSE_FILE services.$peerContainer.environment[3] d  
    yaml w -i $COMPOSE_FILE services.$peerContainer.environment[4] e  
    yaml w -i $COMPOSE_FILE services.$peerContainer.environment[5] f  
    yaml w -i $COMPOSE_FILE services.$peerContainer.environment[6] g  
```
which manually input index is clumsy. script file option might help, but I do not like to use this method 
Is there any plan to support new feature like that
`yaml w -i --append $COMPOSE_FILE services.$peerContainer.environment[]  newValPush2end `

mikefarah:
Yeah totally agree, that would be neat 👍 

I'm time poor atm - but would def accept a feature request

brianreese:
Is there an analog to this feature in 4.x?

PR #39, opened 2017-09-24
Feature: Add append to array

kenjones-cisco:
Adds the ability to append a value to an array instead of having to
know the index of the values within the array.

Resolves: #17


Rambatino:
Is this in the docs? I no can find.

mikefarah:
http://mikefarah.github.io/yq/write/#appending-value-to-an-array-field

Rambatino:
Thanks!

Rambatino:
Can it support like passing a json object? The example is only one element, but I'd like to pass a nested structure to it.

mikefarah:
Only one fields at a time:
```
yq n 'an.array[+].field.list[+].thing' apples
```

```
an:
  array:
  - field:
      list:
      - thing: apples
```

but you could use 
http://mikefarah.github.io/yq/write/#updating-multiple-values-with-a-script
to add multiples

Rambatino:
Thanks @mikefarah!

Issue #25, opened 2017-09-18
List Values Are Unindented


felicianotech:
I can't tell if this is expected, or a bug. Let's say I have the following Yaml file:

```
title: "My Book"
tags:
  - "adventure"
  - "novel"
```

and I run the following command: `yaml write -i mybooks.yml author "J.K. Rowling"

The updated Yaml file now looks like:

```
title: My Book
tags:
- adventure
- novel
author: J.K. Rowling
```

This GitHub Issue is due to the fact that the 'tags' list has lost its indention. That didn't look like valid Yaml to me so I ran it through <http://www.yamllint.com/>. While the Yaml did validate, the `tags` list was "corrected" to the  indention I had it previously. **So is this a bug?**

Also, as you can see, the quotes around the strings are gone. I can see from [this issue](https://github.com/mikefarah/yaml/issues/19) that it's because of the parser so I guess we can't do anything about it. Sucks though.

varac:
Hej, any updates on this ? This behaviour is indeed not great. Yaml lists should be indented although technically they don't have to. I saw the referenced PR #43 got closed, not merged.

mikefarah:
The latest release https://github.com/mikefarah/yq/releases/tag/3.1.0 has a [indent flag](https://mikefarah.gitbook.io/yq/usage/output-format#indent), if this is set to 4 or greater then the lists will be indented - where as an indent of 2 is not.

Not sure why this is the case - this is what the underlying [yaml parser](https://github.com/go-yaml/yaml/tree/v3) does...

ringerc:
An unrelated `yq` tool https://kislyuk.github.io/yq/ has the `--indentless-lists` option. This would be handy to have when working with 4-space indented yaml without list indents to minimise diffs.

Not a biggie as it's possible to use a yaml pretty-printer on it after.

PR #43, opened 2017-09-26
Feature: Adds formatting of yaml output

kenjones-cisco:
The format produced by `gopkg.in/yaml.v2` produces output while valid
is not easy to read especially for arrays.
Adds formatter to handle formatting the marshalled yaml content.
Applies similar logic used by `encoding/json.Indent`.

Resolves: #25


kenjones-cisco:
Not going to work, would have to end up copying most if not all of encoder.go from go-yaml. As such, just going to close and wait for updates to go-yaml to happen.

PR #120, opened 2018-02-16
Support building snap packages for Linux distros

blissd:
Snaps are a new packaging format for Linux distributions. They originated on Ubuntu, but support multiple Linux distributions. It would be wonderful if snap packages could be published with new releases of `yq`, like with brew packages. There is a link to the snap publishing page at the end of the PR.

This PR adds a `snap/snapcraft.yaml` file, allowing a snap package to be build by running:

```
make snap
```
This make target requires that `snapcraft` is installed. On Ubuntu, it can be installed by `snap install snapcraft --classic`, and on MacOS it can be installed by `brew install snapcraft`. Please note I've only tested this on Ubuntu.

If the snap package is published to the Ubuntu Snap Store, then users can install with:

```
snap install yq
```

The `yq` snap package uses `strict` confinement, meaning the `yq` snap has restricted access to the host system. A `home` plug has been added, allowing the `yq` snap package to access files in the users home directory. The `yq` command, when installed as a snap, will not be able to read files outside of the user's home directory.

Useful links:

Building Go snaps: https://docs.snapcraft.io/build-snaps/go

Publishing a snap package: https://docs.snapcraft.io/build-snaps/publish

You may want to publish release to the `stable` channel, and periodically publish the `HEAD` of the master branch to the `candidate` channel. This will allow users to choose which between stable and pre-release versions of `yq`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,9 @@ vendor/*/
 tmp/
 cover/
 yq
+
+# snapcraft
+parts/
+prime/
+.snapcraft/
+yq*.snap
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ help:
 	@echo '    make build           Build yq binary.'
 	@echo '    make install         Install yq.'
 	@echo '    make xcompile        Build cross-compiled binaries of yq.'
+	@echo '    make snap            Build a snap package of yq.'
 	@echo '    make vendor          Install dependencies using govendor.'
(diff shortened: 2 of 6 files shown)


mikefarah:
Cool I hadn't heard of snap packages :)

mikefarah:
Should be able to snap install now, does it work for you?

blissd:
It worked perfectly, thanks.

PR #130, opened 2018-04-01
Add Dockerfile for building the official CLI container

### LINKED ISSUE #126: Add Dockerfile for building publishable images
Hey there! 

Anyway you'd accept a PR that includes a `Dockerfile` intent on distributing `yq` as a container. My use-case is for CI (Circle CI and Concourse can base jobs on containers).

My guess as to your preferred approach might be to use the current build steps to build the binary, and copy it into an Alpine container.

spawnia:
Closes #126

This builds a minimal docker image from sources. It is suitable for use both as a oneshot command as well as an interactive shell.

To make full use of this, an automated build to Docker Hub should be added. Let me know if you need help with setting this up.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM golang:1.9 as builder
+
+RUN go get -u github.com/kardianos/govendor
+
+WORKDIR /go/src/mikefarah/yq
+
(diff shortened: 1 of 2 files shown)


mikefarah:
Awesome! I didn't know about that builder trick - very useful :)

Issue #9, opened 2017-04-26
Can't handle multiple yaml documents within one file


IdanAdar:
Given the following Kubernetes YAML configuration file: 

```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
   name: somename
spec:
   revisionHistoryLimit: 5
   minReadySeconds: 10
   strategy:
      type: RollingUpdate
      rollingUpdate:
         maxUnavailable: 0
         maxSurge: 1
   replicas: 1
   template:
      metadata:
         labels:
            microservice: somename
      spec: 
         containers:
         - name: serviceapi
           image: myimage:0.0.6
           ports:
           - containerPort: 8080
           env:
##### ENVIRONMENT VARIABLES
            - name: somename
              valueFrom:
                configMapKeyRef:
                   name: serviceapi-environment-configuration
                   key: somekey
---
apiVersion: v1
kind: Service
metadata:
  name: somename
  labels: 
     microservice: somename
spec:
   type: NodePort
   ports:
   - port: 8080
   selector:
      microservice: somename
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: some-name
  namespace: default
data:
  someData: data
```

This command will work: 

```
./yaml w -i myfile.yaml spec.template.spec.containers[0].image myimage:0.0.19
```

The `image` value will be replaced, however, the file will then change to this:

```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
   name: somename
spec:
   revisionHistoryLimit: 5
   minReadySeconds: 10
   strategy:
      type: RollingUpdate
      rollingUpdate:
         maxUnavailable: 0
         maxSurge: 1
   replicas: 1
   template:
      metadata:
         labels:
            microservice: somename
      spec: 
         containers:
         - name: serviceapi
           image: myimage:0.0.19
           ports:
           - containerPort: 8080
           env:
##### ENVIRONMENT VARIABLES
            - name: somename
              valueFrom:
                configMapKeyRef:
                   name: serviceapi-environment-configuration
                   key: somekey
```

As you can see, the Services and ConfigMap sections (denoted with `---`) have disappeared.

mikefarah:
Currently this uses go-yaml to parse yaml files and that does not yet support multiple yaml documents embedded into a separate file (https://github.com/go-yaml/yaml/issues/232) :(

IdanAdar:
Too bad... hopefully this will be added - soon!

IdanAdar:
[...]

Issue #16, opened 2017-07-25
Can't handle multiple yaml documents within one file


Godley:
Say I have:
```
---
mykey: myvalue
---
mykey: myothervalue
---
mykey: myotherothervalue
```
I'd expect the tool to be able to handle all three documents and give me a list of values when I run
`yaml r myfile.yaml mykey`, instead it'd just give me the first value.

Is this something you'd be willing to add as a feature? Happy to work on it myself but I have limited go knowledge.

mikefarah:
Yep this is a known issue with the yaml library I'm using https://github.com/mikefarah/yaml/issues/9

Godley:
ah, apologies for the duplication!

mikefarah:
All good - the title on #9 isn't very clear now that I think of it, I'm going to use yours.

svnpenn:
https://github.com/mikefarah/yq/issues/144

svnpenn:
https://github.com/mikefarah/yq/issues/147#

svnpenn:
https://github.com/mikefarah/yq/pull/137

PR #137, opened 2018-05-02
Feature: `yaml read` can read subdocuments beyond the first

therealplato:
This is a first crack at making yq deal with yaml files containing more than one subdocument (https://github.com/mikefarah/yq/issues/9)

```
 ⎇ multiple-documents yq/examples Ω cat multidocument.yaml
---
a: Document One
b:
  c: 1
---
a: Document Two
b:
  c: 2

 ⎇ multiple-documents yq/examples Ω yq read multidocument.yaml a
Document One
 ⎇ multiple-documents yq/examples Ω yq read multidocument.yaml @1.a
Document Two
```

Outstanding API questions:
- I picked `@0, @1.b` to reference the first document and key `b` of the second document.
- Is `@1` a common enough top level key in yaml documents that it would break unacceptable amounts of code?
- Should we zero index or one index? 
  - When you ask for document 2, and there are 2 documents, the current behavior is to error (due to zero indexing.)
- In general, when you ask for a document that's not there, should we just return nil, as if you'd asked for a key that wasn't there?
- What's a clean way to refactor my additions to `func read()` so we can still `defer os.Remove(tmp.Name())`?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ parts/
 prime/
 .snapcraft/
 yq*.snap
+*.swp
--- a/examples/multidocument.yaml
+++ b/examples/multidocument.yaml
@@ -1,0 +1,9 @@
+---
+a: Document One
+b:
+  c: 1
+---
+a: Document Two
+b:
+  c: 2
+
--- a/utils_test.go
+++ b/utils_test.go
@@ -66,6 +66,12 @@ func assertAnyErr(t *testing.T, actualValue error) {
 	}
 }
 
+func assertNilErr(t *testing.T, actualValue error) {
+	if actualValue != nil {
+		t.Error("Expected nil error, got ", actualValue)
+	}
+}
+
 func writeTempYamlFile(content string) string {
 	tmpfile, _ := ioutil.TempFile("", "testyaml")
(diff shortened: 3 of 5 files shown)


therealplato:
I think in case of conflict we override, `--subdoc="---"`, leaving us free to choose a default with a good balance of usability, memorability, not-breakingness, length. I developed against go-yaml/yaml master. Didnt explore their codebase looking for new api

therealplato:
No strong feelings yet on whether placing the document tag as a prepended argument would be better or worse

mikefarah:
I'd prefer to keep the document argument separate - should make things simpler from an implementation point of view, and the usage stays consistent in terms of avoiding conflicts with the top level attributes.

svnpenn:
https://github.com/mikefarah/yq/issues/147

svnpenn:
https://github.com/mikefarah/yq/issues/144

svnpenn:
https://github.com/mikefarah/yq/issues/16

mikefarah:
Ok going to implement this myself - I've been reading up on go-yaml, should be able to update yq to use their Encoder/Decoder streams to handle multiple documents.

PR #138, opened 2018-05-04
Adding a delete node command

### LINKED ISSUE #18: delete node at a path
Is there possible to support write key with empty value like that:
```
networks:  
    networkName:
```

mhuxtable:
Implemented a `delete` command and associated tests for deleting a node from YAML input. This behaves in a similar manner as the `write` command.

The following functionality is not provided, although could conceivably be added with a modicum of additional effort if there is strong opinion that they are required:

 - Specify the node(s) to delete using a script (as with the `write` command)
 - Deleting common elements from all members of an array (deletions from individual members are supported using a path like `a.b[0].name`, but deleting from all instances in one pass, e.g. `a.b[*].name`, is not).

This may close #18, depending on opinion as to whether this is sufficient for providing the required  capability described by that issue.

### DIFF
--- a/data_navigator.go
+++ b/data_navigator.go
@@ -18,9 +18,7 @@ func entryInSlice(context yaml.MapSlice, key interface{}) *yaml.MapItem {
 	return nil
 }
 
-func writeMap(context interface{}, paths []string, value interface{}) yaml.MapSlice {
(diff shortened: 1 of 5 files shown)


mikefarah:
nice!

PR #149, opened 2018-06-06
Add debian packaging

rmescandon:
I consider interesting having the project also available as debian package and here is the required stuff to support it.

I've tested it in a private repo at
```
https://launchpad.net/~rmescandon/+archive/ubuntu/yq
```
You can install the resulting package like this:
```
sudo add-apt-repository ppa:rmescandon/yq
sudo apt update
sudo apt install yq -y
```

I've set myself as Maintainer into debian/control as that's a required metadata. I would like to set Mike Farah as the maintainer but a mail address is required and I haven't found any. Please, Mike, let me know one to fill that field.

I haven't completed any additional related documentation until not having clear in which ppa is going to reside the final .deb package. Ideally that should be one under mikefarah in launchpad, but I offer mine's if you don't want to support it. (Obviously, bumping deb version to 1.15-0)


mikefarah:
This is great, thank you!

I'd prefer to leave you as the maintainer for the debian distro - as I don't use debian and so it would be a bit cumbersome for me to ensure it's working correctly etc (and I have enough on my plate atm).

Happy to merge this in, do you want me to add those instructions to the relevant docos or did you need to change anything else?

rmescandon:
Don't land this yet. I'll complete the documentation and maybe I'll try to make the deployment of the deb a little bit automatic (not sure if that's possible, maybe not).
On the other hand, the Maintainer in the debian/control I bet should be the one maintaining the project, not the packaging :) . That's why I asked you to put there your mail. Though it's fine for me set mine's. If I get some support required maybe I forward the request to you, because I haven't payed much more attention in the internal code.
I'll ping you when ready again to re-review. Thanks

rmescandon:
Sorry for the delay.
I've updated README with the install option from .deb. Launchpad is not letting me distribute if for pure Debian, or at least I haven't seen how to do it, so it's only available for the last two Ubuntu LTS versions. But I think it is ok.
Into release instructions I've left how to publish to a ppa, though in case there is a new version, just mail me and I can quickly update mine's.
Ready to go now.
Thanks.

PR #165, opened 2018-07-14
Travis: Add ppc64le

nitkon:


mikefarah:
Gox is used to cross compile not Travis :)

Issue #180, opened 2018-10-04
Merging in-place changes original file permissions


hossainemruz:
I am merging two file `elasticsearch.yaml` and `custom-config.yaml` using this command.
```
yq merge -i --overwrite elasticsearch.yaml custom-config.yaml
```
Original permission of `elasticsearch.yaml` changes after the merge command.

Permissions before merge command:
```
-rw-rw-r--    1 root     root          1606 Oct  3 09:39 elasticsearch.yml
```
Permissions after merge command:
```
-rw-------    1 root     root          1408 Oct  4 08:09 elasticsearch.yml
```

PR #187, opened 2018-10-24
Fix for Issue #180

thadc23:
A quick fix to copy the original YAML file permissions to the temp file created during an inline merge.


mikefarah:
Awesome work 👍

PR #202, opened 2019-01-20
Bump debian pkg to 2.2-1 version

rmescandon:
Bumped debian pkg to version 2.2-1

Binaries are already published in the repo


mikefarah:
Cheers @rmescandon !

PR #242, opened 2019-05-14
add help description

georgicodes:
👋 Hi there, first time user of `yq` and noticed that there was no description of what yq actually did when you ran the command 😅which meant I had to google to find out. Figured this would help new users of the command.

This PR:
- adds in a short + long description of the command which displays when you run `yq` or `yq --help` + corresponding test case
- updated the README with the help output
- generate new docs as per command


mikefarah:
awesome thanks :)

PR #269, opened 2019-08-23
add Move func to avoid 'invalid cross-device link'

inspirit1986:
Find Move func to avoid this bug. Now merge run without bugs. Please test!

### DIFF
--- a/yq.go
+++ b/yq.go
@@ -37,11 +37,8 @@ func main() {
 	}
 }
 
-/*
-   GoLang: os.Rename() give error "invalid cross-device link" for Docker container with Volumes.
-   MoveFile(source, destination) will work moving file between folders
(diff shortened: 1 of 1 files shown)


mikefarah:
Thanks for the pull request @inspirit1986 - however I actually have a code block for this already, perhaps there's a bug in that *scratches head*. 

If you look on the lines after os.Rename it attempts to copy the file, as it assumes the volume mounting problem you're getting.

I think the issue is after copying the file it's not removing the temporary file, I just need your line

```
 err = os.Remove(sourcePath)
```

After copyFileContents is complete.

If you could fix that I'd be more than happy to merge it in ...

inspirit1986:
Ok, check please.

PR #280, opened 2019-10-04
Issue 216

### LINKED ISSUE #216: [Feature Request] Lint a YAML File
Malformed YAML will crash the parser(#174).

If `yq` support `lint` features that to check whether YAML given is malformed or not, we can use `yq lint` to check YAML before using `yq` to manipulate it.

`yq lint` can work like this:

```bash
$ yq lint correct.yml
ok

$ yq l malformed.yaml
error
```

thadc23:
fixes #216 

Just uses the `yaml.v2` package to attempt to unmarshal the given file into a `map[string]interface{}`.  Naively fails linting if the unmarshal fails.


yq.go
```diff
@@ -241,6 +242,41 @@ Note that if you set both flags only overwrite will take effect.
...
+		path = args[0]
+	}
+	obj := make(map[string]interface{})
+	bytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return err
+	}
+	err = yaml.Unmarshal(bytes, &obj)
+	if err != nil {
+		return errors.New("Failed lint")
```
kenjones-cisco, line 274: why not return the actual error which should provide information about why the yaml file is not valid?

yq.go
```diff
@@ -241,6 +242,41 @@ Note that if you set both flags only overwrite will take effect.
...
+}
+
+func lint(cmd *cobra.Command, args []string) error {
+	var path = ""
+	if len(args) < 1 {
+		return errors.New("Must provide filename")
+	} else if len(args) >= 1 {
+		path = args[0]
+	}
+	obj := make(map[string]interface{})
```
mikefarah, line 267: Yaml files are not necessarily map[string]interface{}, you can also have arrays:
```
- 1
- 2
```

single values

```
cat
```

multi-doc
```
a: thing
---
- 1
- 2
```

mikefarah: You may be better off reusing/refactoring the readProperty function, currently if you don't give it a path it will just spit out the original yaml. Maybe you could extract out the reading functionality it so the lint command runs the same underlying extracted function, but does not print the contents or give a path, and instead just prints out parsing errors if there are any and 'ok' otherwise...

thadc23: That makes sense.  Let me see what I can do.  Thanks!

yq.go
thadc23: Guessing this will go away with using `readProperty` and capturing errors

mikefarah:
No longer accepting new features on v2 - v3 is in progress!

PR #292, opened 2019-11-03
Stop treating top-level values specially

koba-e964:
Before applying this change,
```
$ go build
$ ./yq r - <<<'"true"'
```
returns `true`, which is probably an unexpected behavior (because YAML treats `true` specially.)
After applying this change, the command above returns `"true"`, which is identical to the input YAML.

### DIFF
--- a/yq.go
+++ b/yq.go
@@ -600,12 +600,7 @@ func toString(context interface{}) (string, error) {
 }
 
 func yamlToString(context interface{}) (string, error) {
-	switch context := context.(type) {
-	case string:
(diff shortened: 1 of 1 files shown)


mikefarah:
Nice - but some tests are failing...

mikefarah:
Looks like this wont do multiline strings properly anymore :(

```
./yq r examples/multiline-text.yaml test
```

outputs
```
"abcdefg    \nhijklmno\n"
```

Instead of
```
abcdefg
hijklmno

```

PR #306, opened 2019-12-04
Extract yqlib package and interfaces

cjnosal:
I'm interested in including yq functionality as a module in my own CLI. This requires moving files out of the `main` package and making exportable structs. Additionally, to allow for mocks in unit tests interfaces are also needed.

This PR migrates code to `pkg/yqlib` and `pkg/marshal`, defines new interfaces, and decouples the path/value parsing and write commands from cobra and i/o in `main`. There's no functional change to the code, though some tests have been adapted to test against the new interfaces instead of the private implementation.

I haven't verified the docker/virtualbox scripts (WSL makes that a challenge...) but I've run
`./scripts/test.sh && ./scripts/acceptance.sh && ./scripts/check.sh && ./scripts/coverage.sh && ./scripts/format.sh`


mikefarah:
Cool I like that idea - I'll take a look soon

mikefarah:
This is really nice - thank you!

PR #311, opened 2019-12-13
Chocolatey package

chillum:
Hi Mike,

This PR includes package spec for Chocolatey, as per our email discussion.

To build the package on Linux:
* I'm not sure if you need PowerShell at all, but I think you need mono and chocolatey for this, my google-fu gave only [this link](https://stackoverflow.com/questions/39386471/can-i-build-chocolatey-packages-on-linux)
* navigate to `choco` directory and execute something like `mono /path/to/choco.exe pack yq.nuspec`. it will either work, or not. if not, consider changing the `tools` path in the `.nuspec` file (see comment).
* when it works, it will output `yq.2.4.1.nupkg` file, send it to me for verification to be sure everything is okay

after we're done with it, basically two more things to mention:

1. publishing a package. sign up on [chocolatey community packages](https://chocolatey.org/packages) and basically just submit the `yq.2.4.1.nupkg` file through web UI. there's also a CLI helper for that step (`choco push`).
2. if you will release a new version, change `version` in `.nuspec`, `Url`, `Url64`, `Url64bit` and `Checksum64` in `.ps1`. then pack, then push.

I believe, that's it. Drop me a note with the resulting `.nupkg` and whether I need to update `tools` path in the `.nuspec` (I'll update the PR if you won't do it already).

P.S. this PR does not document `choco` installation yet, let's get a package submitted and only then document this way of installing.

### DIFF
--- a/choco/tools/chocolateyinstall.ps1
+++ b/choco/tools/chocolateyinstall.ps1
@@ -1,0 +1,14 @@
+﻿$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+
+$packageArgs = @{
+    PackageName    = 'yq'
+    Url            = 'https://github.com/mikefarah/yq/releases/download/2.4.1/yq_windows_386.exe'
+    Checksum       = '7ED72DF0565DC08481101DC5B5E00FC071791189541803F71BDD1BEB7FE90522'
+    ChecksumType   = 'sha256'
+    Url64bit       = 'https://github.com/mikefarah/yq/releases/download/2.4.1/yq_windows_amd64.exe'
+    Checksum64     = 'BDFD2A00BAB3D8171EDF57AAF4E9A2F7D0395E7A36D42B07F0E35503C00292A3'
+    ChecksumType64 = 'sha256'
+    FileFullPath   = "$toolsDir\yq.exe"
+}
+
+Get-ChocolateyWebFile @packageArgs
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ parts/
 prime/
 .snapcraft/
 yq*.snap
+
+# choco
+*.nupkg
--- a/choco/yq.nuspec
+++ b/choco/yq.nuspec
@@ -1,0 +1,27 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>yq</id>
+    <version>2.4.1</version>
+    <packageSourceUrl>https://github.com/mikefarah/yq/tree/master/choco</packageSourceUrl>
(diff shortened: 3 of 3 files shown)


imsergiobernal:
+1 ! choco power 👏

mikefarah:
Neato! I'll take a look at this soon in the new year :)

choco/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,14 @@
+﻿$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+
+$packageArgs = @{
+    PackageName    = 'yq'
+    Url            = 'https://github.com/mikefarah/yq/releases/download/2.4.1/yq_windows_386.exe'
```
rhockenbury, line 5: Would it be possible to bundle these executables in the package rather than downloading from the web to support those that have air-gapped environments?

chillum: I believe, the download way is preferred by Choco guidelines.

every Choco package installed on my system uses downloads, so I believe Choco is pretty useless without Internet connection. they do have some kind of 'Choco server for your local net' or something like this that they market for use by business, I believe it's the answer for air-gapped environments for some reason not allowing connections to GitHub.

mikefarah:
I've had a go at trying to get this to work on my mac - but it's too tricky.

I think it'd be better off for someone else to maintain the chocolate package - I'd be happy to put a link/instructions to the choco installation... @chillum

chillum:
@mikefarah yeah, I can put this into my repo with choco packages and submit that.

but maybe somebody of more regular yq users want to take it (I admit I might not check for updates often)?

I think, if nobody will chime in until Sunday, I'll go ahead and submit it. does this sound like a plan?

mikefarah:
Sounds reasonable to me...

chillum:
okay, went ahead and submitted 2.4.1, after it's accepted will submit v3.
the new home for Choco package will be https://github.com/chillum/choco-packages/tree/master/yq

after that all happens, will notify that everything is working.

mikefarah:
Cool - give me instructions when its ready and I'll update the docs

030:
@chillum Could you indicate when the package will be published to chocolatey.org?

vasily-codefresh:
@030 sure I will. the package passed automated testing on 2 Feb, still awaiting for human review to be approved. this might take time, I recall my other packages waited almost 2 months. =(

030:
@mikefarah Could you indicate what information could be added to start the review?

chillum:
okay, [chocolatey package for yq](https://chocolatey.org/packages/yq/) is approved, currently there're submitted versions 2.4.1 and 3.2.1.

`choco install yq` for installation.

choco package source lives in [my repo](https://github.com/chillum/choco-packages), PRs for new versions are welcome. we may close this PR.

030:
@mikefarah could you add a link to the readme?

PR #316, opened 2020-01-01
Add handle for int Issue 314

tnimni:
This solves the problem when number with 7 digits or more is written as exponent


tnimni:
@mikefarah Hi, Can you possibly respond here?
I appreciate your help

mikefarah:
Nice - I will try to adapt this in v3

mikefarah:
This has been fixed by another PR

PR #318, opened 2020-01-05
Allow merge commands to specify overwrite and append simultaneously

cjnosal:
Mergo allows specifying `WithOverride` and `WithAppendSlice`, resulting in slices being merged and scalars being replaced. `yqlib.Merge` currently only allows one or the other.


mikefarah:
Thank @cjnosal - I'm actually working on v3 of this which is a significant rewrite, I've taken your changes and ported them across ;)

Issue #314, opened 2019-12-30
yq write numbers as exponent instead integer


tnimni:
When i run a command to replace a value in yaml
for example 
yq w fileName.yaml global.containerTags.site 1212121 > values.new
expected value is
gloabl:
  containerTags:
    site: 1212121

actual value is 1.212121e+06

this only happen if i try to write a number with 7 digits or more 6 digits or less works fine

mikefarah:
This will be fixed in v3

gharris999:
Same as #245

mikefarah:
Fixed in https://github.com/mikefarah/yq/releases/tag/3.0.1

PR #319, opened 2020-01-10
#314 Fix the big int changing into float

RyanSiu1995:


mikefarah:
Thanks @RyanSiu1995  - I've taken these changes and ported them across to the v3 branch

PR #342, opened 2020-02-05
Fix path generation when merging file has period in key

coryrc:
The program generates a path for every leaf node in the
file-to-be-merged. It does not escape them if they contain a dot, as
the path-expressions document mentions is necessary.

Add in a test for this condition. Verified it fails without the fix.

### DIFF
--- a/cmd/commands_test.go
+++ b/cmd/commands_test.go
@@ -1262,7 +1262,7 @@ c:
   toast: leave
   test: 1
   tell: 1
-  taco: cool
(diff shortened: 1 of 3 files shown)


mikefarah:
Awesome! Thanks @coryrc ! I've manually merged it into the compare branch - as that's the branch I'm working on for the next release. Should be out soon!

coryrc:
Hooray! The code is well-architected, so it was easy to find the place to fix it.

Issue #344, opened 2020-02-05
Upgrade to 3.0 not working on Ubuntu?


MaurGi:
I am getting 2.2-1 is the latest release:

```bash
$  sudo apt update && sudo apt upgrade yq -y
Hit:1 https://deb.nodesource.com/node_10.x bionic InRelease
Hit:3 https://packages.microsoft.com/repos/azure-cli bionic InRelease                                                                                                                                                              
Hit:4 https://packages.microsoft.com/ubuntu/16.04/prod xenial InRelease                                                                                                    
Hit:5 http://security.ubuntu.com/ubuntu bionic-security InRelease                                                                                                          
Hit:6 http://security.ubuntu.com/ubuntu xenial-security InRelease                                                                                           
Hit:7 http://archive.ubuntu.com/ubuntu bionic InRelease                                                                                                     
Hit:8 http://ppa.launchpad.net/rmescandon/yq/ubuntu bionic InRelease                          
Hit:2 https://packages.cloud.google.com/apt kubernetes-xenial InRelease 
Hit:9 http://archive.ubuntu.com/ubuntu bionic-updates InRelease         
Hit:10 http://archive.ubuntu.com/ubuntu bionic-backports InRelease
Reading package lists... Done                      
Building dependency tree       
Reading state information... Done
All packages are up to date.
Reading package lists... Done
Building dependency tree       
Reading state information... Done
yq is already the newest version (2.2-1).
Calculating upgrade... Done
The following package was automatically installed and is no longer required:
  libdumbnet1
Use 'sudo apt autoremove' to remove it.
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

mikefarah:
@rmescandon was maintaining the ubunut release - are you able to make a new one? I'll be making another release soon with a number of fixes..

rmescandon:
There's a newer version of the deb package for Eoan matching 3.1.1 version

mikefarah:
Hey @rmescandon there were a few bugs identified with 3.1.1 :embarrassed face: when you get a chance would you be able to bump the version up to 3.1.2?

PR #367, opened 2020-02-24
Bump deb to 3.1 2

rmescandon:
#344 

- bumped deb packaging version to 10
- bumped standard versions to 4.1.4
- updated copyright from propietary to Expat (MIT)
- handle distro dirs by using yq.files 
- exclude docs folder from packaging, to focus only on yq binary


mikefarah:
Thanks again @rmescandon !

mikefarah:
note to self - merge then close

Issue #134, opened 2018-04-17
Colorize the output?


tadayosi:
Is there way to colorize the output of yq on terminal like jq does?

mikefarah:
not yet!

Frodox:
Heavy plus!
Can't find any cli tools for colorize output of yml

pawel-swiecki-saucelabs:
+1

risentveber:
Seems like one should use this https://github.com/goccy/go-yaml#ycat

risentveber:
@tadayosi see my pr please)

mikefarah:
Fixed in https://github.com/mikefarah/yq/releases/tag/3.2.0 (thanks @risentveber !!)

See https://mikefarah.gitbook.io/yq/usage/output-format#colorize-output

PR #376, opened 2020-02-27
Add colorization

risentveber:
Proof of concept for https://github.com/mikefarah/yq/issues/134


mikefarah:
That is super cool!

mikefarah:
Pity it has to reparse the yaml - but it'd be super handy. Just need to see if I can hook it into the other commands output too

mikefarah:
Refactored to work for all commands (see new PR) - thanks again! This will go out in the next release

Issue #403, opened 2020-04-09
Stargazers over time


030:
## Stargazers over time

[![Stargazers over time](https://starchart.cc/mikefarah/yq.svg)](https://starchart.cc/mikefarah/yq)

mikefarah:
hah neat

PR #404, opened 2020-04-09
[GH-403] Stargazers over time

030:


mikefarah:
I like looking at this chart, it gives me a sense of pride :)

But other than contributors - I don't think people will be interested...

PR #416, opened 2020-04-17
docs(readme): protect parameter expansions

mminot-yseop:
Hi.

The Readme’s Bash function example might cause issues (and head-scratching) due to how its parameters are handled.

From `man bash`:

> "$@" is equivalent to "$1" "$2" ...

Example:

```bash
  $ function quoted { printf '%s\n' "$@"; }
# ≠
  $ function unquoted { printf '%s\n' $@; }

  $ quoted 1 '2 3' 4
1
2 3
4

  $ unquoted 1 '2 3' 4
1
2 ← Unwanted split into two words.
3 ↙
4
```

(It gets even worse when people start fiddling with the value of IFS.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,20 +60,20 @@ GO111MODULE=on go get github.com/mikefarah/yq/v3
 Oneshot use:
 
 ```bash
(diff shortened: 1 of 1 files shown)


mikefarah:
Nice

PR #451, opened 2020-06-02
bash completion for yq read; requires jq

amkhlv:
Dear Mike, thank you for this excellent tool ! I wrote a bash TAB-completion script for `yq read`. Would you consider adding it to "utils" section?

### DIFF
--- a/utils/yq-completion.bash
+++ b/utils/yq-completion.bash
@@ -1,0 +1,41 @@
+
+_yq() {
+    local cur prev coms opts car cdr x
+    coms="compare delete help merge new prefix read validate write"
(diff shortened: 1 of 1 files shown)


mikefarah:
Glad you like it - under the hood, for CLI processing, this uses sp13/cobra, and that has a bash completion feature that would be nice to add:

https://github.com/spf13/cobra/blob/master/bash_completions.md

This also has support for zsh and fish.

mikefarah:
I've implemented the functionality out of cobra, thanks for the idea though 👍

mikefarah:
I should add - check out the latest version (3.3.1) and the shell-completion command :)

amkhlv:
thank you !

PR #461, opened 2020-06-12
mk TempDir

runzexia:
in some docker images, such as busybox.
There may not be a tmp directory, so we will try to create it if it does not exist.

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -423,6 +423,14 @@ func readAndUpdate(stdOut io.Writer, inputFile string, updateData updateDataFn) 
 		if err != nil {
 			return err
 		}
+		// mkdir temp dir as some docker images does not have temp dir
+		_, err = os.Stat(os.TempDir())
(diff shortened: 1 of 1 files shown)


cmd/utils.go
```diff
@@ -423,6 +423,14 @@ func readAndUpdate(stdOut io.Writer, inputFile string, updateData updateDataFn)
...
 			return err
 		}
+		// mkdir temp dir as some docker images does not have temp dir
+		_, err = os.Stat(os.TempDir())
+		if os.IsNotExist(err) {
+			err = os.Mkdir(os.TempDir(), 0700)
+			if err != nil {
+				return err
+			}
+		}
```
mikefarah, line 435: nice - but we need to also handle the case when os.Stat returns an error that isn't a 'IsNotExist' error (and return that error)

runzexia: fixed

PR #507, opened 2020-07-30
Debian build improvements

dukelion:
We have our own pipeline for debian packages and configuration for dh_golang in this repo doesn't work for us.
I've made a few minor changes to have it working on broader range of build environments.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-yq (3.3-0) focal; urgency=medium
+yq (3.3.0) focal; urgency=medium
 
   * You can control string styles (quotes) using the new --style flag
   * String values now always have quotes when outputting to json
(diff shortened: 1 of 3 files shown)


debian/changelog
```diff
@@ -1,4 +1,4 @@
-yq (3.3-0) focal; urgency=medium
+yq (3.3.0) focal; urgency=medium
```
dukelion, line 1: > dpkg-source: error: can't build with source format '3.0 (native)': native package version may not have a revision

mikefarah:
I have no strong opinion on this @rmescandon ?

debian/changelog
rmescandon: @dukelion did you try '3.0 (quilt)' format with 3.3-0 version pattern?

rmescandon:
Actually yes. The official releases have the format x.x.x, so let's move to that one.

debian/changelog
dukelion: Yes, it didnt work either.

PR #542, opened 2020-09-17
Add bash support for yq

geekette86:
Due to a need in work, I made this so I install bash so I don't use two images in my pipes or add more steps and lines 
all in one place method ;-)
I don't know if it will be really helpful for others but why not


mikefarah:
As an alternative, you can create your own docker image with yq as a base, then install whatever you like.

```
FROM mikefarah/yq:3
RUN apk update && apk upgrade && apk add bash
```

Not going to merge this in sorry - otherwise I'd end up with all sorts of docker files to manage and release.

geekette86:
actually that is why I did here https://hub.docker.com/repository/docker/maneloyez/yq

Issue #546, opened 2020-09-23
Exploding multiple anchors doesn't result in the final anchor being honoured.


StuWill:
YQ 3.4.0, when exploding multiple anchors, doesn't overwrite values correctly. This doesn't happen using YQ 3.2.1.

Operating system: tested on MacOs and Linux

data1.yml
```yaml
a: &first_anchor
  foo: bar
  first: first
b: &second_anchor 
  foo: bazz
  second: second
c:
  <<: *first_anchor
  <<: *second_anchor
```

When running `yq r data1.yml -X` 
output:
```yaml
a:
  foo: bar
  first: first
b:
  foo: bazz
  second: second
c:
  foo: bar
  first: first
  second: second
```

However the expected output should be (specifically `c.foo` should equal `bazz`):
```yaml
a:
  foo: bar
  first: first
b:
  foo: bazz
  second: second
c:
  foo: bazz
  first: first
  second: second
```

Please ask if you need any additional information, cheers!

combor:
Just to make it more clear - c.foo value should be overwritten and be `bazz`

mikefarah:
Verified, as a work around, you can do array merges:

```
c:
  <<: [*first_anchor, *second_anchor]
```

And the override will work correctly.

combor:
Some parts work ok but other do not. Another test case:
```
a: &first_anchor
  foo: bar
b: &second_anchor 
  foo: bazz
c: &third_anchor
 <<: *first_anchor
d: &fourth_anchor
 <<: *second_anchor
e:
  <<: [*third_anchor, *fourth_anchor]
```
exploding the entire file:
```
yq r -X /tmp/x.yml 
a:
  foo: bar
b:
  foo: bazz
c:
  foo: bar
d:
  foo: bazz
e:
  foo: bazz
```
works OK but reading a specific key do NOT:
```
yq r -X /tmp/x.yml e
{}
```
BUT if we read even more specific key:
```
yq r -X /tmp/x.yml e.foo
bazz
```
It works OK 🤷

combor:
Another test case with more extreme bug :)
```
uk:
  lab_common: &uk_lab_common
    minio_endpoint: http://lab-services-minio:9001
  dev_common: &uk_dev_common
    <<: *uk_lab_common
  dc1_common: &uk_dc1_common
    minio_endpoint: http://dc1-services-minio:9001
  dev4:
    <<: [*uk_dc1_common, *uk_dev_common]
```
```
[...]

PR #551, opened 2020-09-30
Overwrite node if key exists but value doesn't match or append if key does not exist

combor:
This fixes a bug with exploding anchors not overwriting nodes reported
in https://github.com/mikefarah/yq/issues/546

Test case:
```
a: &first_anchor
  foo: bar
  first: first
b: &second_anchor
  foo: bazz
  second: second
c:
  <<: *first_anchor
  <<: *second_anchor
```

Output:

```
a:
  foo: bar
  first: first
b:
  foo: bazz
  second: second
c:
  foo: bazz
  first: first
  second: second
```

foo is properly overwritten by *second_anchor

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -149,10 +149,15 @@ func writeString(writer io.Writer, txt string) error {
 	return errorWriting
 }
 
-func setIfNotThere(node *yaml.Node, key string, value *yaml.Node) {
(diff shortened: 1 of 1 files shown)


combor:
I just realised that this will overwrite any `foo: value` that already exists in the node `c:` and doesn't come from an anchor. This is wrong. Will close

PR #559, opened 2020-10-07
keep order of keys when json marshalling

### LINKED ISSUE #435: JSON conversion (still) does not preserve object order
**Describe the bug**

It seems from #274 having been fixed that conversion to JSON should now be preserving object order, but that's not what I'm seeing.

**Input Yaml**
data1.yml:
```yaml
foo: a
bar: b
```

**Command**
```
yq r -jP data1.yaml
```

**Actual behavior**
```json
{
  "bar": "b",
  "foo": "a"
}
```

**Expected behavior**
```json
{
  "foo": "a",
  "bar": "b"
}
```

**Additional context**
yq version 3.3.0 on macOS 10.15.3

thehowl:
I've implemented a custom solution to #435. (fixes #435)

This implementation uses a custom type, `orderedMap`, with custom marshalers/unmarshalers for both JSON and YAML.

Tests have been changed accordingly to allow for the new behaviour.

### DIFF
--- a/cmd/read_test.go
+++ b/cmd/read_test.go
@@ -1441,3 +1441,23 @@ func TestReadFindValueDeepObjectCmd(t *testing.T) {
 `
 	test.AssertResult(t, expectedOutput, result.Output)
 }
+
+func TestReadKeepsKeyOrderInJson(t *testing.T) {
+	const content = `{
+	"z": "One",
+	"a": 1,
+	"w": ["a", "r"],
+	"u": {"d": "o", "0": 11.5},
+}`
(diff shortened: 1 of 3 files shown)


mikefarah:
Neat! Ill give it a go before merging in

mikefarah:
Looks good, can you add a explicit test for this though? I use the tests as technical documentation of what this is supposed to do - so if it fails I know which part of functionality failed.

Like TestReadKeepsKeyOrderInJson or something

thehowl:
@mikefarah done

PR #611, opened 2020-12-21
#607 Fix string value with spaces error

### LINKED ISSUE #607: Can't write a string value with a white space
**Describe the bug**
I can't write a string value containing a white space.

Version of yq: 4.0.0
Operating system: mac
Installed via: release/homebrew

**Command**
```
yq e -n '.test = "Hello world!"'
```

**Actual behavior**

```
Error: Parsing expression: Lexer error: could not match text starting at 1:9 failing at 1:16.
        unmatched text: "\"Hello "
```

**Expected behavior**

```yaml
test: Hello world!
```

**Additional context**
It works if there is no white space:
```
yq e -n '.test = "Hello-world!"'
```

Thanks for your help

djajcevic:
Fixes #607 
I've added one more test that produces documentation entry.

The reason behind excluding space inside `"..."` is unknown to me, but since all tests are passing I think we are good. If needed, I can go a bit more into implementation to figure out a better solution.

By examining https://github.com/timtadh/lexmachine#adding-a-pattern I found that it is safe to remove that space from the regex exclude group since `lexer.Add([]byte(` `), token("SPACE"))` is not an option for YQ.

### DIFF
--- a/pkg/yqlib/operator_assign_test.go
+++ b/pkg/yqlib/operator_assign_test.go
@@ -13,6 +13,13 @@ var assignOperatorScenarios = []expressionScenario{
 		},
 	},
 	{
+		description: "Create yaml file with value with spaces",
+		expression:  `.a.b = "cat" | .x = "frog jumps"`,
+		expected: []string{
+			"D0, P[], ()::a:\n    b: cat\nx: frog jumps\n",
+		},
+	},
+	{
 		description: "Update node to be the child value",
 		document:    `{a: {b: {g: foof}}}`,
 		expression:  `.a |= .b`,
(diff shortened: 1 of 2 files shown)


mikefarah:
Thanks @djajcevic!

mikefarah:
Just a fyi in case you (hopefully) contribute more, I've moved the test case to operator_value_test.go

djajcevic:
@mikefarah should this be in Milestone 4.1.0 or you are planning to make a bug fix release (4.0.1). 
I'm using a locally compiled version, for now, but would like to see this released as soon as possible.

mikefarah:
Should hopefully release 4.1.0 tmrw along with other fixes, otherwise it will have to wait till after christmas.

djajcevic:
If you want, I can release 4.0.1 :) (if by following instructions you wrote I have permissions)

Issue #619, opened 2020-12-23
`make` fails with "./bin/golangci-lint: No such file or directory"


cswarth:
**Describe the bug**

`make` on the top level yq repoistory fails with `./scripts/check.sh: line 6: ./bin/golangci-lint: No such file or directory`

This happens because Dockerfile.dev invokes `/opt/devtools.sh` in its home directory which is "/go".
This installs golangci-lint into /go/bin.

Later the tool cannot be found when invoked with `docker run -w /yq/src/github.com/mikefarah/yq`

Anyone building from source should see the same error so I'm not sure why I seem to be the only one to have run into this.

Version of yq: V4 branch or master as of sha eb539ff3267e18abc7c8257d30010911a92f4e73
Operating system: mac
Installed via: not installed, building

**Actual behavior**

```
$ make
Sending build context to Docker daemon  9.308MB

Step 1/6 : FROM golang:1.15
 ---> 5f9d35ce5cfe
Step 2/6 : COPY scripts/devtools.sh /opt/devtools.sh
 ---> Using cache
 ---> 14404a9334de
Step 3/6 : RUN set -e -x     && /opt/devtools.sh
 ---> Running in 288cc93c0794
[91m+ /opt/devtools.sh
[0m[91m+ go get golang.org/x/tools/cmd/goimports
[0m[91m+ + shwget -s v1.24.0 -O-
 -nv https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh
[0m[91m2020-12-23 06:06:07 URL:https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh [10739/10739] -> "-" [1]
[0m[91mgolangci/golangci-lint info checking GitHub for tag 'v1.24.0'
[0m[91mgolangci/golangci-lint info found version: 1.24.0 for v1.24.0/linux/amd64
[0m[91mgolangci/golangci-lint info installed ./bin/golangci-lint
[0mRemoving intermediate container 288cc93c0794

[... many lines of output...]

docker run --rm -e LDFLAGS=" -X main.GitCommit=eb539ff+CHANGES -X main.GitDescribe=v4.1.0-2-geb539ff" -e GITHUB_TOKEN="" -v /Users/cswarth/src/yq/vendor:/go/src -v /Users/cswarth/src/yq:/yq/src/github.com/mikefarah/yq -w /yq/src/github.com/mikefarah/yq yq_dev bash ./scripts/check.sh
[...]

PR #620, opened 2020-12-23
find golangci_lint through PATH

cswarth:
Addresses issue #619 - make fails....

Adds /go/bin to PATH env variable in Dockerfile.dev and modifies  scripts/check.sh to search PATH instead of running golangci-lint directly via relative path.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -4,6 +4,7 @@ COPY scripts/devtools.sh /opt/devtools.sh
 
 RUN set -e -x \
     && /opt/devtools.sh
+ENV PATH=/go/bin:$PATH
 
 # install mkdocs
 RUN set -ex \
--- a/scripts/check.sh
+++ b/scripts/check.sh
@@ -3,7 +3,7 @@
 set -o errexit
 set -o pipefail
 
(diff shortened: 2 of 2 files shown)


mikefarah:
Must admit I personally don't use docker for building, thanks for investigating and fixing though! 

I can confirm this fixes the docker build for me, but it broke my local build :( looking into it - I'm guessing its a local machine setup thing..

mikefarah:
Yeah running locally, normally you'd run ./scripts/devtools.sh from the source directory, which places the tool under ./bin.

Hmm perhaps updating ./scripts/check.sh to check whether the executable is in the relative directory, failing that try PATH

cswarth:
So you just use `go build` to build?   I was pleasantly surprised that docker build was available so I didn't have to go through the hassle of installing all the tools just to try it out.

Sorry for breaking local build.  Want me to try to fix it or are you taking care of this?

mikefarah:
I use make local build

If you could that'd great, otherwise ill get around to it eventually :)
________________________________

mikefarah:
Manually merged in (and fixed to work locally and in docker)

Issue #640, opened 2021-01-04
Document or add V4 Yqlib.Get usage equivalent


mkatychev:
**Document or add example of using yqlib on preexisting yaml nodes.**

I was curious if there is a practical way to pass yaml nodes to `Yqlib` in the same way `Yqlib.Get` works:
```go
Yqlib.Get(rootNode *yaml.Node, path string) ([]*NodeContext, error)
```

Note:
- All the public methods/functions I have found expect a byte slice, string, or filepath
- There does not seem (at least in the time I've spent trying to replicate the functionality) to be a pre-decoded ingress for the path traversal flow by passing in a 

**Describe the solution you'd like**
Provide an example similar to the `Get` method that returns a `*CandidateNode`
**Describe alternatives you've considered**

**Additional context**
My current project uses pre-decoded `yaml.Node`s that it passes to yqlib v3, the flow doesn't seem to match for v4

mikefarah:
I could add a new function to the all_once_evaluator, bsically it would be doing

```
func EvaluateNodes(expression string, inputCandidates *list.List) (*list.List, error) {
   node, err := treeCreator.ParsePath(expression)
   return = treeNavigator.GetMatchingNodes(input, node)
}
```

mkatychev:
I created a PR https://github.com/mikefarah/yq/pull/657 to try and implement the suggestion.

mikefarah:
Fixed in 4.3.2

Note that I ended up moving your code @mkatychev to the Evaluator interface

mkatychev:
Awesome, thanks for the quick turnaround!

PR #657, opened 2021-01-12
added EvaluateNodes and EvaluateCandidateNodes to yqlib

mkatychev:
In an attempt to reproduce v3 yqlib's  `Get`
```go
Yqlib.Get(rootNode *yaml.Node, path string) ([]*NodeContext, error)
```

Two functions have been added:
* `func EvaluateNodes(expression string, nodes ...*yaml.Node) (*list.List, error)`
* `func EvaluateCandidateNodes(expression string, inputCandidates *list.List) (*list.List, error)`

This attempts to address the feature suggested in https://github.com/mikefarah/yq/issues/640

### DIFF
--- a/pkg/yqlib/lib.go
+++ b/pkg/yqlib/lib.go
@@ -166,6 +166,24 @@ func NodeToString(node *CandidateNode) string {
 	return fmt.Sprintf(`D%v, P%v, (%v)::%v`, node.Document, node.Path, tag, buf.String())
 }
 
+// EvaluateNodes takes an expression and one or more yaml nodes, returning a list of matching candidate nodes
+func EvaluateNodes(expression string, nodes ...*yaml.Node) (*list.List, error) {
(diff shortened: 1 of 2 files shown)


pkg/yqlib/lib.go
```diff
@@ -166,6 +166,24 @@ func NodeToString(node *CandidateNode) string {
 	return fmt.Sprintf(`D%v, P%v, (%v)::%v`, node.Document, node.Path, tag, buf.String())
 }
 
+// EvaluateNodes takes an expression and one or more yaml nodes, returning a list of matching candidate nodes
+func EvaluateNodes(expression string, nodes ...*yaml.Node) (*list.List, error) {
```
mkatychev, line 170: Not sure what kind of testing pattern should go in `lib_testing.go` since these functions are not compatible with the `expressionScenario` pattern

PR #755, opened 2021-03-22
Add chocolatey badges

adriens:
Add chocolatey badges so windows user immediatly know they can install ys easily


mikefarah:
Thanks @adriens - I've put the chocolate badge near the chocolate installation method - as its a community supported installation method

PR #865, opened 2021-06-20
Consistancy in docker image tag

pratikbalar:
Stay consistent with docker image tag as per release version for good practice and standardization.


mikefarah:
I don't really like this change tbh - we'll end up with `mikefarah/yq:4` vs `mikefarah/yq:v4.9.6`. I think for docker tags it's idiomatic not to have the leading `v`.

Where as for `go` packages you usually do have a leading `v`.

PR #873, opened 2021-06-29
Update debian pkg files to version 4.9.6

rmescandon:
Sorry for the delay on this. 
Binaries are alreay deployed on ppa for Focal (Ubuntu 20.04)


mikefarah:
👍🏼

PR #937, opened 2021-09-12
supports multiline output

### LINKED ISSUE #936: Multiline outputs support in the GitHub action

- Support multiline outputs in the Github Actions
- Even though `yq` produces a multiline output, only the first line of it is stored in the `outputs.result`
- Current behaviour:
  example: https://github.com/pjxiao/yq-action/runs/3577428434

  ```yaml
  jobs:
    convert-to-json:
      runs-on: ubuntu-latest
      steps:
        - uses: actions/checkout@v2
        - name: This outputs multiline JSON
          uses: mikefarah/yq@v4.12.2
          id: act
          with:
            cmd: yq eval --output-format=json 'config.yml'
  
        - name: But we get only the first line here
          env:
            ACT_RESULT: ${{ steps.act.outputs.result }}  # = "{"
          run: echo "${ACT_RESULT}"
  ```

- If `yq` produces a multiline output, the entire output is store in the `outputs.result`
- Expected behaviour:

  example: https://github.com/pjxiao/yq-action/runs/3577428423

  ```yaml
  jobs:
    convert-to-json:
      runs-on: ubuntu-latest
      steps:
        - uses: actions/checkout@v2
        - name: This outputs multiline JSON
          uses: mikefarah/yq@v4.12.2
          id: act
          with:
            cmd: yq eval --output-format=json 'config.yml'
  
        - name: I'd like to get the entire JSON
          env:
            ACT_RESULT: ${{ steps.act.outputs.result }}
          run: echo "${ACT_RESULT}"
  ```

**Additional context**

- This is useful when we use `yq` as a YAML/JSON converter

pjxiao:
- Fixes #936
- Supports multiline output

Thank you so much for your work maintaining yq

### DIFF
--- a/github-action/entrypoint.sh
+++ b/github-action/entrypoint.sh
@@ -2,5 +2,8 @@
 set -e
 echo "::debug::\$cmd: $1"
 RESULT=$(eval "$1")
+RESULT="${RESULT//'%'/'%25'}"
+RESULT="${RESULT//$'\n'/'%0A'}"
+RESULT="${RESULT//$'\r'/'%0D'}"
 echo "::debug::\$RESULT: $RESULT"
 echo ::set-output name=result::"$RESULT"


github-action/entrypoint.sh
```diff
@@ -2,5 +2,8 @@
 set -e
 echo "::debug::\$cmd: $1"
 RESULT=$(eval "$1")
+RESULT="${RESULT//'%'/'%25'}"
+RESULT="${RESULT//$'\n'/'%0A'}"
+RESULT="${RESULT//$'\r'/'%0D'}"
```
pjxiao, line 7: This escape logic comes from [a Github support case](https://github.community/t/set-output-truncates-multiline-strings/16852/3)

mikefarah:
Cool thanks for the PR - happy to merge this in, do you mind explaining how this works first though?

github-action/entrypoint.sh
mikefarah: bash black magic 👍🏼

pjxiao:
@mikefarah I appreciate your reply.

Suppose we have the following workflow:

```yaml
jobs:
  convert-to-json:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: This outputs multiline JSON
        uses: mikefarah/yq@v4.12.2
        id: yaml-to-json
        with:
          cmd: yq eval --output-format=json 'config.yml'

      - run: echo "${{ steps.yaml-to-json.outputs.result }}"  # Currently we get only the first line
```

In the example above, we try to convert a YAML to a mutiline JSON but `outputs.result` contains only the first line.
example: https://github.com/pjxiao/yq-action/runs/3577428434 (please see 'Only the first line is saved' step)

My patch allows us to get an entire output even if the output contains line-breaks.
example: https://github.com/pjxiao/yq-action/runs/3577428423 (please see 'The entire JSON is saved' step)

The escaping logic:
The added escaping logic just URL encodes `%`, CR and LF. This originates in [a community post](https://github.community/t/set-output-truncates-multiline-strings/16852/3). Though I researched documentations and references [1], I couldn't find any mentions of ecaping. So this behaviour is an internal specification.

[1]: https://docs.github.com/en/actions/

PR #996, opened 2021-11-08
Setup dependabot for github-actions, docker and go dependencies 

mmorel-35:
First of all thank you for yq, it's a great tool!
This PR updates the current dependencies with the help of dependabot.


mikefarah:
I didn't know you could do that with github, neat!

PR #997, opened 2021-11-09
setup golangci-lint-action

mmorel-35:
Hi !

Golangci-lint can be used directly as an action.
It allows you to see the warning and errors directly in the files of the PR.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -7,10 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.15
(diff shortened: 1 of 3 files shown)


mikefarah:
Thanks @mmorel-35  - I'd prefer to stick to a single build check - otherwise I've got two spots to keep in sync, the make file and the CI config.

That said, I'm going to take your idea of running the checks in PRs and update the one I've got to do that.

mmorel-35:
If you want to keep using the script only way, you can still integrate the ".golangci.yml" to manage it's configuration.

PR #1026, opened 2021-11-30
Add support for Podman as well as Docker

rhatdan:
Would like to see this package work on systems without Docker installed
which have Podman.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>


mikefarah:
Hi @rhatdan  - happy to add this support in. With docker - it get built and published as part of the github action release build. This is why users can run `mikefarah/yq`. Does podman have it's own repository or will it just work with the docker repo too?

rhatdan:
Podman Supports multiple repositories, and does not hard code docker.io, but fully supports it.
The is no podman.io repository.

Issue #596, opened 2020-12-08
Ubuntu PPA is missing arm64 builds


qmfrederik:
The package repository for Ubuntu contains amd64 and i686 builds of yq, but not arm64 (nor armhf) builds.

It would be awesome if you could enable the PPA to enable arm64 (and potentially armhf) builds, too. This would allow us to use the PPA on Raspberry Pi devices, among others.

robertomier:
from now on arm64 is included.

PR #1046, opened 2021-12-21
Update changelog for version v4.16.2

robertomier:
arm64 build is also configured in the ppa builds, as per requested in #596

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -3,28 +3,20 @@ yq (4.16.2) focal; urgency=medium
   * Fixed with semicolon space issue
   * Updating with documentation
   * Added STDIN example to the top
-  * Added STDIN example to the top
-  * Bump deb to version 4.13.0
   * minor readme cleanup
   * Help text tweak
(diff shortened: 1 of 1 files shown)


mikefarah:
it's prob worth de-duping some of those lines

robertomier:
> it's prob worth de-duping some of those lines

Absolutely. They were the comments extracted from the master commits. I've removed the non-sense or duplicated ones. Let me know if some others should be also removed.

In any case, I'd say that there should be some kind of control over comments in commits. They should include atomic changes (features, documentation, fixes, etc..) and be representative (I see some recent commits whose comment is simply "wip" :/)
Do you think worths following [these rules](https://www.conventionalcommits.org/en/v1.0.0/)?

mikefarah:
You could just use the release notes https://github.com/mikefarah/yq/blob/master/release_notes.txt

Yeah those wip comments I was supposed to rebase them away on the branch...oops

PR #1075, opened 2022-01-21
fixes path in scripts/generate-man-page-md.sh

mr-pmillz:
I noticed that `./pkg/yqlib/doc/headers/Main.md` is no longer a valid path, appears to be missing operators in path.

This PR should fix that 😃 

Awesome project BTW


mikefarah:
Thanks good pickup!

PR #1082, opened 2022-01-25
Add a new `envsubst` operator to replace environment variables in strings

sciyoshi:
## Motivation

The `env` operator is great for reading in and replacing individual environment variables, but for some use cases it would be useful to be able to replace `${ENVVAR}` within strings automatically. One can already do this using the standard `envsubst` command, but this has all the normal pitfalls of naive string templating with yaml files. With yq, this becomes much safer and more powerful.

## Description

Adds an `envsubst` operator which works on strings and uses the [envsubst](https://github.com/a8m/envsubst) go library, which provides niceties like fallback values (e.g. `${ENVVAR:-fallback}`).

One possibility I considered was making `envsubst` thread across mappings and lists as well. This makes envsubst more flexible and possibly easier to use, but also makes it less orthogonal since this can also be achieved with `(.. | select(tag == "!str")) |= envsubst`. I added a note about this to the docs, but it may be another useful addition to consider, since it would make a yaml-safe envsubst replacement as simple as doing `yq e envsubst -`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/mikefarah/yq/v4
 
 require (
+	github.com/a8m/envsubst v1.2.0
 	github.com/elliotchance/orderedmap v1.4.0
 	github.com/fatih/color v1.13.0
 	github.com/goccy/go-yaml v1.9.5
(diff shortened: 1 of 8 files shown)


mikefarah:
Super cool! You're the first contributor too add a new operator in - you win this emoticon 🥇

PR #1108, opened 2022-02-13
Fix zsh completion

whi-tw:
4ec533b introduced a bug which causes the 'hidden' cli parameter `__complete` to be ignored.

Add a check for this parameter, so that it can be correctly passed to `cobra`

There is still an issue - `yq <tab>` will suggest that a subcommand is run, not a file. I think 'pick eval by default' could be done in a more cobra-friendly way, but I just don't know enough about it to implement that

### DIFF
--- a/acceptance_tests/completion.sh
+++ b/acceptance_tests/completion.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+testCompletionRuns() {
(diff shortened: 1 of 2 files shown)


mikefarah:
Yeah I couldn't find a way to do it in cobra either - thanks for the fix.

PR #1146, opened 2022-03-19
update cli help

justin-f-perez:
Hi! json -> (idiomatic) yml conversion is the main (only) reason I installed yq but I tend to forget the syntax. I created a "json2yml" alias for myself but I thought this updated `--help` might be useful to new users. While the `--input-format` flag states "Note that json is a subset of yaml," (and I feel a little dumb admitting it) it was non-obvious to me that this meant no `--input-format` is required to parse json.

- remove duplicate comment
- add common use case (json -> idiomatic yml)


mikefarah:
Note that you can just run `yq -P thing.json` too

PR #1171, opened 2022-04-08
Added action to automatically prepare a branch for the alpine package update

Roemer:
This change adds a GitHub Action that automatically creates a branch over at https://gitlab.alpinelinux.org/ (in your own fork) with all changes needed for a new version of the Alpine Package to be published.
All that is left then is quickly check the branch and create a merge request back to the upstream (although the merge request could also be automatically created via the gitlab api if we want that).

This PR helps keeping the alpine package up to date automatically (except for the final merge request, which can be automated as well if we want).

Hope to get some feedback what you think about that. Happy to do adjustments if wanted.


mikefarah:
This is awesome! Would you be able to also provide the code for automatically creating the merge request - but commented out? That way, once I give it a run and check it all works, I can easily turn it on.

Also the MR has failed the build - I've got a number of linters and gosec running...

Looking forward to give it a whirl!

Roemer:
Yes I can add the code and enable/disable it with a flag. Will check the errors also. Will try to do this tomorrow.

mikefarah:
Thanks @Roemer - I'll see if I can fix the remaining issues and get it to work!

Roemer:
Ok cool. Feel free to come back if you have issues/questions.

mikefarah:
I had a try - the problem is go doesn't like having multiple projects embedded into a single repo.

The linters complain - particularly gosec.

This makes me realise that this code would be better as a separately published github action that _anyone_ can use to publish to alpine... thoughts? I'd be keen to use that...

Roemer:
With go 1.18 there is a new workspace feature for multi-workspaces. Maybe that helps? One alternative would be to just run this script in a separate repository. In general, it should be possible to create a github action out of this script, I kept it fairly general with variables all over the place, only a few places where I hardcoded some things like `yq` which would need to be changed. Also I'm not totally sure how flexible the editing of the `APKBUILD` needs to be.

PR #1222, opened 2022-05-23
Fix some typos in docs

creydr:
Fixes some typos in the docs

### DIFF
--- a/how-it-works.md
+++ b/how-it-works.md
@@ -47,7 +47,7 @@ b: dog
 
 ## Complex assignment, operator precedence rules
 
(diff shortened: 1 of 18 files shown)


mikefarah:
Thanks for this! The operator docs are actually generated from test files, so they'll need to be updated too otherwise your changes will be clobbered. Happy to merge in if you update those too - otherwise I can find the spots and update them when I get a chance.

creydr:
> Thanks for this! The operator docs are actually generated from test files, so they'll need to be updated too otherwise your changes will be clobbered. Happy to merge in if you update those too - otherwise I can find the spots and update them when I get a chance.

:+1: updated it

PR #1239, opened 2022-06-14
XML decoder additions

rndmit:
Add flag for keeping XML-attribute namespace (e.g. "xmlns:xsi")
Add flag for disabling translation namespace prefixes to their corresponding URLs

### DIFF
--- a/cmd/constant.go
+++ b/cmd/constant.go
@@ -11,6 +11,8 @@ var inputFormat = "yaml"
 var xmlAttributePrefix = "+"
 var xmlContentName = "+content"
 var xmlStrictMode = false
+var xmlKeepNamespace = false
+var xmlUseRawToken = false
 
 var exitStatus = false
 var forceColor = false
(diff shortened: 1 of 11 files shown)


mikefarah:
This looks really cool - I'll have to have an in depth look when I get a chance

mikefarah:
I'm thinking that the defaults for these new flags should actually be true 🤔  that way, the default behavior is to persist the existing namespaces - what do you think?

rndmit:
I think it's a good idea to hold namespaces by default. But not sure about defaulting Raw Token usage. 
On the one hand, it doesn't resolve namespaces to URLs and for me, it looks preferred. But the [doc](https://pkg.go.dev/encoding/xml#Decoder.RawToken) says that it also "_does not verify that start and end elements match_". So this could be a problem for some users. And I also didn't find other ways to disable this translation.

rndmit:
E.g. parsing a broken file with Token
`<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
</aproject>`
will cause an error. But parsing this with RawToken successful, and also fixes missing closing tag (`</aproject> -> </project>`).

mikefarah:
Yeah I think that's OK for default behaviour. The flags are there for strict mode, turning off raw mode etc. I think more often than not people will want to maintain their namespaces. I will release soon - within the week I'm hoping :)

rndmit:
Hopefully no one relies yq to verify XML validity 😄
Can't wait to use it in my CIs

P.S. And also thank you for such a great tool! I like it a lot

PR #1241, opened 2022-06-16
adding support for --wrapScalar=false in properties encoder

dcarbone:
I'm currently running into a scenario where it would be useful to wrap the value side of the output from the Properties encoder.

This PR is a first draft at that with some tests, please let me know if you'd like some more!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ yq.1
 # debian pkg
 _build
 debian/files
+
+# intellij
+/.idea
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -93,7 +93,7 @@ func configureEncoder(format yqlib.PrinterOutputFormat) yqlib.Encoder {
 	case yqlib.JSONOutputFormat:
 		return yqlib.NewJONEncoder(indent, colorsEnabled)
 	case yqlib.PropsOutputFormat:
-		return yqlib.NewPropertiesEncoder()
+		return yqlib.NewPropertiesEncoder(unwrapScalar)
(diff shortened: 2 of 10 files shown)


mikefarah:
This looks really good - you just need to add test(s) into properties_test.go, this will also generate documentation so others know about the feature 👍🏼

dcarbone:
@mikefarah i've updated the test workflow a bit and added one for my added feature.

if you disagree with the changes, i can work on a different solution.

dcarbone:
@mikefarah i've updated the test workflow a bit and added one for my added feature.

if you disagree with the changes, i can work on a different solution.

mikefarah:
Thank you! There is a problem though in the documentation generated, it doesn't show the updated command to get the wrapping (e..g supplying `--unwrapScalar=false` - if you have a look at xml_test.go, you can see that there are added 'scenarioType's that generate a slightly different documentation on the command, as well us parameterising the encoder/decoder correctly.

dcarbone:
@mikefarah sorry it took a few days to get back to you on this!  i reverted my initial pass as it wasn't really in keeping with your original design, and i didn't want to start imposing my own opinions there.

with this update, i've done the following:

1. replaced the various `if .. else ..` cases when processing scenarios with a switch
  * this was mostly to make it a smidge cleaner to read and expand upon, and also forces future `scenarioTypes` to have an explicit handler, rather than letting it succeed with ambiguous correctness.
1. replaced a few `var` definitions with `const` as their values are indeed constant
1. added a new scenario type (`encode-wrapped`) for the properties test suite.

let me know what you think :)

Issue #1256, opened 2022-06-24
panic when mutiple assign to multi documents


care0717:
**Describe the bug**
I've got panic when I use [multiple](https://mikefarah.gitbook.io/yq/operators/multiply-merge) assign to multi documents.

Version of yq: 4.25.2
Operating system: mac
Installed via: homebrew

**Input Yaml**
sample.yml:
```yaml
a: 1
---
b: 2
```

**Command**
```
yq '. *= {"c":3}' sample.yaml
```

**Actual behavior**
output
```
a: 1
c: 3
panic: interface conversion: interface {} is nil, not yqlib.multiplyPreferences

goroutine 1 [running]:
github.com/mikefarah/yq/v4/pkg/yqlib.multiplyOperator(0xc000013890?, {0xc0004612f0, 0xc000461380, 0x1, {0x0, 0x0}}, 0xc0003b8c30)
	github.com/mikefarah/yq/v4/pkg/yqlib/operator_multiply.go:37 +0x205
github.com/mikefarah/yq/v4/pkg/yqlib.(*dataTreeNavigator).GetMatchingNodes(0xc000461380?, {0xc0004612f0, 0xc000461380, 0x1, {0x0, 0x0}}, 0xc0003b8c30)
	github.com/mikefarah/yq/v4/pkg/yqlib/data_tree_navigator.go:37 +0x25c
github.com/mikefarah/yq/v4/pkg/yqlib.resultsForRHS(0x19?, {0xc0004612f0, 0xc000461380, 0x1, {0x0, 0x0}}, 0x28?, {0x0, 0x0, 0xc00045e920}, ...)
	github.com/mikefarah/yq/v4/pkg/yqlib/operators.go:69 +0x1fc
github.com/mikefarah/yq/v4/pkg/yqlib.doCrossFunc(0x1d?, {0xc0004612f0, 0xc000461380, 0x1, {0x0, 0x0}}, 0xc0003b8c48, {0x0, 0x0, 0xc00045e920})
	github.com/mikefarah/yq/v4/pkg/yqlib/operators.go:125 +0x413
github.com/mikefarah/yq/v4/pkg/yqlib.crossFunctionWithPrefs(0x1?, {0xc000461260, 0x0, 0x1, {0x0, 0x0}}, 0x1e00108?, {0x0, 0x0, 0xc00045e920})
	github.com/mikefarah/yq/v4/pkg/yqlib/operators.go:158 +0x3db
[...]

PR #1257, opened 2022-06-24
Fix panic when multiply assign

care0717:
I fix https://github.com/mikefarah/yq/issues/1256

When multi documents exist, I've got panic because of `expressionNode.Operation.Preferences = nil`.
But I've got type assertion error instead of this issue when I remove it. So I make type assertion safe.

### DIFF
--- a/pkg/yqlib/operator_assign.go
+++ b/pkg/yqlib/operator_assign.go
@@ -22,8 +22,8 @@ func assignUpdateOperator(d *dataTreeNavigator, context Context, expressionNode 
 	}
 
 	prefs := assignPreferences{}
-	if expressionNode.Operation.Preferences != nil {
-		prefs = expressionNode.Operation.Preferences.(assignPreferences)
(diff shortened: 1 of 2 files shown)


care0717:
I'm wondering what kind of test cases to add.

Before I fix it, this test case is passed.
```
	{
		skipDoc:    true,
		document:   "a: 1\n---\nb: 2",
		expression: `. *= {"c": 3}`,
		expected: []string{
			"D0, P[], (!!map)::a: 1\nc: 3\n",
			"D1, P[], (!!map)::b: 2\nc: 3\n",
		},
	},
```
Unlike stream_evaluator, operator_test push all node to inputList before GetMatchingNodes. (maybe I think)

Let me know what you think.

mikefarah:
Wow this is a really good find! You are right about stream_evaluator vs operator_test as well. I think the best way to test this case (for now) is to add an acceptance test. I'm happy to do that.

PR #1267, opened 2022-07-06
Fix flag

karlhorky:
Hi @mikefarah , thanks for yq!

I wanted to fix the flag here, since it seems that the correct flag should be `--output-format` instead of `--outputformat` on [the Working with CSV, TSV page](https://mikefarah.gitbook.io/yq/usage/csv-tsv)


karlhorky:
@mikefarah Thanks for the merge!

PR #1293, opened 2022-08-04
add build argument to use different uid/gid

yvolchkov:
While uid/gid 1000 will work in most setups, sometimes users have
different uid/gid. For example is a shared machine with multiple
users.


mikefarah:
Cool - though can't you just adjust the user when running the image https://github.com/mikefarah/yq#running-as-root ?

yvolchkov:
rigth. Seems I was a bit overthinking. The `root` does cover my case

PR #1303, opened 2022-08-13
Document dynamic get and output in GitHub Action

MattiSG:
Thanks for `yq` and its GitHub Action integration!

I encountered a not-so-easy limitation: GitHub Actions escape and interpolate rules can conflict with some yq syntax. In particular, it took me some time to figure out how to quote a variable that could contain dots in a query path (more usage context: https://github.com/ambanum/OpenTermsArchive/pull/899).

This changeset adds such an example to the README. I wasn't sure if you'd prefer to have it there or in the `gitbook` branch so I went for the easiest :wink:


mikefarah:
Awesome thanks for this, I've add it to gitbook as well  👍🏼

PR #1317, opened 2022-08-26
Removing USER directive from the github action Dockerfile

mollonado:
The problem that this tries to solve is enabling in-place editing of yaml files that are the result of the checkout github action. The files are checkout in the GITHUB_WORKSPACE directory by the runner user.

As mentioned here - https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#docker-container-filesystem we should omit the USER directive from the Dockerfile. 

The problem is twofold: 
- the runner user is with uid 1000 and here we set the USER to 1001
- the documentation mentions to not set the USER for the github action container.

In the main yq container we already set the USER to yq which has uid 1000, which currently matches the runner user uid. So this will kind of work as it is but only if we do not set the USER to 1001 in the github-action/Dockerfile.

### DIFF
--- a/github-action/Dockerfile
+++ b/github-action/Dockerfile
@@ -2,7 +2,4 @@ FROM mikefarah/yq:4
 
 COPY entrypoint.sh /entrypoint.sh
 
-# this seems to be the default user in github actions
-USER 1001
-
 ENTRYPOINT ["/entrypoint.sh"]


mikefarah:
According to that documentation, it needs to run under root - which the yq dockerfile does not do (it was a change made some time ago, a recommended security practice https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user, https://sysdig.com/blog/dockerfile-best-practices/)

The USER I added in the github action was specifically to work with the checkout container - I have a test specifically for this scenario and it still passes: https://github.com/mikefarah/yq/actions/workflows/test-yq.yml. Adding that user was to solve a bug against writing in place, it's what fixed the problem. 

This Docker USER thing is a bit of a pain - I'm pretty sure if it's removed from the action the test will start failing again.

PR #1383, opened 2022-10-17
Fix set-output deprecate; keys are case sensitive

mattphelps-8451:
One line fix to resolve #1382 for https://github.com/guihkx and other currently broken downstream workflows.

@MattiSG @mikefarah

PR #1390, opened 2022-10-18
Fix set-output deprecate; multiline was broken

mattphelps-8451:
This is a followup to #1383 and closes the loop on set-output deprecation issues our team was still having using master yq. 

We were using yq for merging base values with env specific values in Helm. This commit has been tested on our end and fixes multiline usage of yq by the following:

1. Remove escaping of special characters
    -  Otherwise you'll encounter breaking changes for example `appVersion: "1.0.0"%0Aapplica...` where `%0A` is escaped newline `\n` but shouldn't be escaped.
    -  Read more here: https://github.com/orgs/community/discussions/26288#discussioncomment-3876281

2. Delimiter is required for multiline strings
    -  https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
    -  Delimeter must be random. UUID is a solid choice. I verified that `/proc/sys/kernel/random/uuid` does indeed exist on the yq image.

@mikefarah @MattiSG 

----

### Example of broken workflows before:
Screen Shot 2022-10-18 at 4 55 18 PM

### After:
Screen Shot 2022-10-18 at 4 56 12 PM

### DIFF
--- a/github-action/entrypoint.sh
+++ b/github-action/entrypoint.sh
@@ -2,10 +2,11 @@
 set -e
 echo "::debug::\$cmd: $1"
 RESULT=$(eval "$1")
-RESULT="${RESULT//'%'/'%25'}"
-RESULT="${RESULT//$'\n'/'%0A'}"
-RESULT="${RESULT//$'\r'/'%0D'}"
 echo "::debug::\$RESULT: $RESULT"
 # updating from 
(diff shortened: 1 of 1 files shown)


mikefarah:
Thanks for you work!

PR #1418, opened 2022-11-09
fix lib.go unhandled error

washanhanzi:
Unfortunately, I can't find a way to reproduce this bug with a minimal csv file.

When I try to convert a CSV file to JSON.
The `err` returned from `decoder.Decode()` method is: `"yaml: found character that cannot start any token"`.
And this makes `parsedNode` `nil`, and `parsedNode.Node` will panic.

So I added a line to check the returned `err` from `Decode()` method.


mikefarah:
Oops my mistake - thanks for raising a PR!

PR #1439, opened 2022-11-18
Draft: Toml

mikefarah:


mikefarah:
Hi @pelletier - I've got started on trying to write a decoder using your API - I'm a little confused as to how the table types work?

pkg/yqlib/decoder_toml.go
```diff
@@ -0,0 +1,223 @@
...
+	err = dec.d.DeeplyAssign(context, path, valueNode)
+	if err != nil {
+		return nil, err
+	}
+
+	return rootMap.Node, nil
+}
+
+func (dec *tomlDecoder) createTable(tomlNode *toml.Node) (*yaml.Node, error) {
+	log.Debug("Table: %v", string(tomlNode.Data))
```
mikefarah: here specifically - what's the toml.Node structure for the table? If you look at the test I've got in toml_test.go - I have
```
[owners]
blah = "X"
```

But I only see to get owners in the node child here 🤔

pelletier: Correct. The parser returns top level expressions. So you should have a first node for the table header, then another one for key-value.

pelletier: So in this example you need two calls to NextExpression() to parse the whole document (arguably 3 to detect that you're at the end of the document and stop).

PR #1454, opened 2022-11-30
Improve release-deb.sh

RadxaYuntian:
* Enable binary package build in addition to source package (both will be generated)
* Fix a bug in the `build.sh` where host $PATH overwrites docker $PATH and could break build (for example when running from Arch host)
* Fix build when host user does not have UID 1000 (for example GitHub runner's UID is 1001)
* Allow cross build for other architecture with `--platform` flag (currently support `x86_64` and `aarch64` only)
* Also add CI script to test this code in every push

Our company builds custom Debian images for our hardware products. As such we prefer to have a proper deb distributed via our apt repo, instead of wgetting a random binary from GitHub release page. We also build all our packages with GitHub workflows, and they have to run on ARM64 machines, so this PR added those capabilities.

Since `yq` is an important dependency for us, we also added a basic CI workflow to make sure the packaging script will not go bad without being noticed.

### DIFF
--- a/scripts/release-deb.sh
+++ b/scripts/release-deb.sh
@@ -17,6 +17,7 @@ VERSION=
 DISTRIBUTION=
 DO_SIGN=
 PASSPHRASE=
+PLATFORM="$(uname -m)"
 
 show_help() {
   echo "  usage: $(basename "$0") VERSION [options...]"
@@ -39,6 +40,7 @@ show_help() {
   echo "    --ppa PPA                   Push resultant files to indicated ppa. This option should be given along with a signing key."
   echo "                                  Otherwise, the server could reject the package building. Default is set to 'rmescandon/yq'"
   echo "    --passphrase PASSPHRASE     Passphrase to decrypt the signage key"
+  echo "    --platform PLATFORM         Targeting platform. Default to $(uname -m)"
   exit 1
 }
 # read input args
@@ -83,6 +85,10 @@ while [ $# -ne 0 ]; do
       shift
       PASSPHRASE="$1"
       ;;
+    --platform)
+      shift
+      PLATFORM="$1"
+      ;;
     *)
       if [ -z "$VERSION" ]; then
         VERSION="$1"
@@ -94,6 +100,19 @@ while [ $# -ne 0 ]; do
 done
 
 [ -n "$VERSION" ] || (echo "error - you have to provide a version" && show_help)
+
+case $PLATFORM in
+  x86_64)
+    PLATFORM="amd64"
+    ;;
+  aarch64)
+    PLATFORM="arm64"
+    ;;
+  *)
+    echo "error - unsupported platform $PLATFORM "
+    exit 1
+    ;;
+esac
 
 if [ -n "$OUTPUT" ]; then
   OUTPUT="$(realpath "$OUTPUT")"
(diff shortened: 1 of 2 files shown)


mikefarah:
Hi - I don't actually look after the deb releases - @rmescandon does (though he hasn't been active for a while).

That said, if we can automate it, I'd be happy to include it in the build - the problem is until I know what the script does and how - I'm not willing to merge it in. I'll presumably also need to get some sort of debian something account so I can sign and publish these packages. Not sure what's involved. There's a lot of black magic in this I don't understand - not the least being:

```
"sed -i -E \"s/^(yq):(x?):([0-9]+):([0-9]+):(.*):(.*):(.*)$/\1:\2:$(id -u):$(id -g):\5:\6:\7/\" /etc/passwd 
```

RadxaYuntian:
Right now the CI only does the build but not publishing, so the deb release can be handled in the same way today.

> I'll presumably also need to get some sort of debian something account so I can sign and publish these packages. 

The package doesn't strictly require signing. It is the Release file (which is like the primary index file) that requires signing. This one contains the hash for Packages file, which in turn contains the hash for debs. However, all those files only makes sense in an apt repo, so if you plan to publish it in GitHub release, it is no different than the unsigned binaries that we currently have. Debian system will happily install an unsigned deb without warning.

The sed command in question matches yq's [passwd line](https://linux.die.net/man/5/passwd) which contains 7 fields. Then it substitutes the UID and GID field with the current calling user's same value, while keep the rest fields untouched. This is to have the yq user inside the docker image to have the same UID and GID as the callee who likely owns the source code.

`release-deb.sh` calls `gbp dch` within the docker, which updates `debian/changelog`. This requires yq user inside the docker to have write permission on the source code. However, when mounting the host folder in docker image, there is no mapping of UID or GID. So if the host user doesn't belong to the 1000:1000 as the yq user defaults to (ex. GitHub runner or in a shared build machine which we also have), this user cannot touch that file and build fails.

There are a few different ways to workaround this issue:

1. Update the file permission from the host side.
  Build scripts shouldn't change user's code, and the host user may need `sudo` permission to update the file permission, which may not be possible in an automated build environment or a public build machine. 
  The resulting build output uses yq user's UID/GID, and definitely requires `sudo` to remove them.
2. Run as root in docker.
  Less bad but the build output now belongs to root, so still requires sudo from host.
3. Fix yq user UID/GID.
  This is what I did in this PR. The result is no more file permission issues and the user can manipulate the build artifacts just like a native build.

I hope this explains why the seeming magic sed command is needed.

mikefarah:
I don't actually like that the original script created a change log from the git commits - I'd prefer it used the release_notes.txt file. Is the sed black magic required because of release notes? I'd much prefer to manually commit/copy the release notes somewhere and simplify the build - as part of the release I already manually update the release notes anway...

RadxaYuntian:
This is also how we manage the change log. We use git commit message as a starting point to update `debian/changelog`, then packaging script will just use what's there instead of coming up their own. I could make such adjustment and see if that's enough to drop the sed magic.

mikefarah:
Hey @RadxaYuntian not sure if you are still interested in this. FYI rmescandon is no longer maintaining the debian build, so (for now) I have removed those instructions from the readme...

RadxaYuntian:
In that case I think we can rework the current Debian package infrastructure and get rid of his hacks. I'll revisit this later.

scripts/release-deb.sh
```diff
@@ -228,37 +247,33 @@ RUN apt-get -qq -y --no-install-recommends install \
...
     rsync \
-    sensible-utils && \
+    sensible-utils \
+    sudo && \
   apt-get clean && \
   rm -rf /tmp/* /var/tmp/*
 
 COPY --from=golang /usr/local/go /usr/local/go
 
 # build debian package as yq user
```
rossigee, line 257: Is that really necessary? Why not just build it as an existing user, such as 'nobody'?

RadxaYuntian: That was coded by previous author so I can't answer why. But my guess is that the script needs write permission in mapped folder, so you cannot use nobody.

RadxaYuntian:
This is currently abandoned, as Debian 12 packaged a different [`yq`](https://salsa.debian.org/python-team/packages/yq).

PR #1547, opened 2023-02-05
Adding instructions to install on Arch Linux

JoaoGFarias:


mikefarah:
Hi - this is actually in the Readme already, under "Community Supported Installation methods" - because its not an installation method I manage myself :)

Issue #1526, opened 2023-01-18
Shell string encoder


vit-zikmund:
**Please describe your feature request.**
Hi Mike (thanks for this great tool and remarkable development effort), it would be great to have something as `jq`'s `@sh` operator that encodes a string into a literal shell string representation, so one could safely generate things like `.env` files and otherwise combine `yq` with shell scripts.

**Describe the solution you'd like**
In general, one should be able to wrap a string with single quotes and escape any that are in the payload. The handling differs from the yaml `"single"` quoting style, though.

Let's say we have this little `test.yaml`:
```yaml
a: |-
  some string with spaces,
  newline and ' quote '
```

So the result from `yq '.a | @sh' test.yaml` would do the same as:
```sh
$ yq . test.yaml -o json | jq -r '.a | @sh'
'some string with spaces,
newline and '\'' quote '\'''
```
Shell considers any consecutive string of character as a single token as long as it doesn't contain unescaped/unquoted characters from the special `IFS` env variable, which by default contains `<space><tab><newline>`. These are the default (and usually only) separators. I would safely assume this to be the case to not make the code overly complicated. As there's no way how to escape a single quote within a single-quoted string, one needs to exit the current quoted block, print a literal escaped quote and start another quoted block, which is exactly what can be seen in the output above.

Naturally, this encoder would only apply to the string type and should bail out on another type. Also I think there's no need to have a decoder thanks to all the environment variable handling operators `yq` already has.

**Describe alternatives you've considered**
I've imagined that some folks would also appreciate this to be a single-line output. Although this is not supported in lightweight shells, the widespread BASH can use literal escapes in a "dollar single quoted string" like `$'\n'`, which is a simple newline. So in addition to the above, the encode could escape newlines with `\n` and wrapping that part with `$'...'` instead of plain `'...'`.

In that case the output from `yq '.a | @sh("oneline")' test.yaml` could look like this:
```sh
# when blindly wrapping all '-separated parts with `$'...'
$'some string with spaces,\nnewline and '\'$' quote '\'$''
# or when using `$'...' only for parts containing a newline
$'some string with spaces,\nnewline and '\'' quote '\'''
# or escaping just the newlines same as the quotes
'some string with spaces,'$'\n''newline and '\'' quote '\'''
```
I don't have any preference of the above, each gets the job done, however the first and last can be implemented as a single pass without holding a context. Also while this alternative is a fun-to-have, it's way beyond the original desire ;)

[...]

PR #1548, opened 2023-02-06
Use a lazy-quoting @sh encoder

vit-zikmund:
I eventually took a hit at the second alternative from https://github.com/mikefarah/yq/issues/1526#issuecomment-1413650693, which is generally ouptut-size optimal and behaves nice for strings with a lot of safe characters.

### DIFF
--- a/pkg/yqlib/doc/operators/encode-decode.md
+++ b/pkg/yqlib/doc/operators/encode-decode.md
@@ -478,7 +478,7 @@ yq '.coolData | @sh' sample.yml
 ```
 will output
 ```yaml
-'strings with spaces and a \'quote\''
+strings' with spaces and a '\'quote\'
 ```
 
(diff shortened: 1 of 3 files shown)


vit-zikmund:
I also tried to figure out how to code in a quoting style switch the user could use, but I see this is not really supported for `@encoder` style calls. So for now I added an internal switch.

mikefarah:
Looks great, thanks for the contribution :)

mikefarah:
It would be good to add tests for the empty quote cases you mentioned. Happy to do that myself if you don't have the time

vit-zikmund:
I'll do it in a couple minutes ;)

vit-zikmund:
Huh, it took a little more :face_exhaling: Starting a yaml string starting with a single quote that's in a go string literal... It's good that I like the movie Inception :smile:

PR #1550, opened 2023-02-08
Add ``--nul-output|-0`` flag to separate element with NUL character

vaab:
This option is provided to ensure solid parsing of complex data (with any binary content except NUL chars) by separating the `yq` root collection member's output with NUL char. As a safe-guard, an error will be cast if trying to use NUL character with content that contains itself NUL characters inside.

I'm aware of the possible heated debate on the inclusion of this as it happened over a similar request on https://github.com/stedolan/jq/issues/1271 ... As a heavy bash user, IMHO, NUL separated output is the safest way to treat complex data in bash, and allows to pipe results directly with many other tools without mangling in bash. The previous thread however have raised important concerns about NUL chars that could be appear in the valid output of a query and could potentially mess with unprepared code badly. 

This is why I added a check when using `-0` option to fail with an error when the data being separated with NUL is containing a NUL characters itself. In practice, this is nearly never expected and if it happens, I felt I wanted to be in position to fail and warn the user.

I did write some acceptance tests and am very welcome to any of your suggestions. I'll write unit test if you feel this PR is worthwhile for you.

Please note that to be sure of NOT triggering the error, the flag `-r=false` with the yaml encoder needs to be added and NUL chars will be encoded in yaml format which doesn't include raw NUL chars. But the better would probably to fail in the calling code (which is made possible with this implementation) when encountering data that is not formatted as you had expected.

As a side-note, I'm the author of `shyaml`, and am using it extensively in bash projects, and would be happy to archive my project : it is slow. Alas, without the current PR, `yq` is not able to replace current usages of `shyaml` in my case.

There are workaround that I'm aware of (mainly from the `jq` thread posted previously) that are too costly in term of efficiency, code readability or simple data solidity in my own opinion.

I would totally understand if you are not interested in this addition on your side. The magic of open-source allows me to keep using it on my side. I'm curious to receive your feedback, as these are complex topics.

This project is probably my first time fiddling with Go language and project, so I probably did some newcomer's mistake. I'll be happy to correct them if you can stress them out for me.

Many thanks for writing `yq`, that's a really awesome project.

### DIFF
--- a/acceptance_tests/nul-separator.sh
+++ b/acceptance_tests/nul-separator.sh
@@ -1,0 +1,286 @@
+#!/bin/bash
+
+setUp() {
+    rm test*.yml || true
+}
+
+## Convenient bash shortcut to read records of NUL separated values
+## from stdin the safe way. See example usage in the next tests.
+read-0() {
+    local eof="" IFS=''
+    while [ "$1" ]; do
+        ## - The `-r` avoids bad surprise with '\n' and other interpreted
+        ##   sequences that can be read.
+        ## - The `-d ''` is the (strange?) way to refer to NUL delimiter.
+        ## - The `--` is how to avoid unpleasant surprises if your
+        ##   "$1" starts with "-" (minus) sign. This protection also
+        ##   will produce a readable error if you want to try to start
+        ##   your variable names with a "-".
+        read -r -d '' -- "$1" || eof=1
+        shift
+    done
+    [ -z "$eof" ] ## fail on EOF
+}
+
+## Convenient bash shortcut to be used with the next function `p-err`
+## to read NUL separated values the safe way AND catch any errors from
+## the process creating the stream of NUL separated data.  See example
+## usage in the tests.
+read-0-err() {
+    local ret="$1" eof="" idx=0 last=
+    read -r -- "${ret?}" <<<"0"
+    shift
+    while [ "$1" ]; do
+        last=$idx
+        read -r -d '' -- "$1" || {
+            ## Put this last value in ${!ret}
+            eof="$1"
+            read -r -- "$ret" <<<"${!eof}"
+            break
+        }
+        ((idx++))
+        shift
+    done
+    [ -z "$eof" ] || {
+        if [ "$last" != 0 ]; then
+            ## Uhoh, we have no idea if the errorlevel of the internal
+            ## command was properly delimited with a NUL char, and
+            ## anyway something went really wrong at least about the
+            ## number of fields separated by NUL char and the one
+            ## expected.
+            echo "Error: read-0-err couldn't fill all value $ret = '${!ret}', '$eof', '${!eof}'" >&2
+            read -r -- "$ret" <<<"not-enough-values"
+        else
+            if ! [[ "${!ret}" =~ ^[0-9]+$ && "${!ret}" -ge 0 && "${!ret}" -le 127 ]]; then
+                ## This could happen if you don't use `p-err` wrapper,
+                ## or used stdout in unexpected ways in your inner
+                ## command.
+                echo "Error: last value is not a number, did you finish with an errorlevel ?" >&2
+                read -r -- "$ret" <<<"last-value-not-a-number"
+            fi
+        fi
+        false
+    }
+}
+
+## Simply runs command given as argument and adds errorlevel in the
+## standard output. Is expected to be used in tandem with
+## `read-0-err`.
+p-err() {
+    local exp="$1"
+    "$@"
+    printf "%s" "$?"
+}
+
+wyq-r() {
+    local exp="$1"
+    ./yq e -0 -r=false "$1"
+    printf "%s" "$?"
+}
+
+testBasicUsageRaw() {
+  cat >test.yml <<EOL
+a: foo
+b: bar
+EOL
+
+  printf "foo\0bar\0" > expected.out
+
+  ## We need to compare binary content here. We have to filter the compared
+  ## content through a representation that gets rid of NUL chars but accurately
+  ## transcribe the content.
+  ## Also as it would be nice to have a pretty output in case the test fails,
+  ## we use here 'hd': a widely available shortcut to 'hexdump' that will
+  ## pretty-print any binary to it's hexadecimal representation.
+  ##
+  ## Note that the standard `assertEquals` compare its arguments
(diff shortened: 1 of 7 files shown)


mikefarah:
Hi @vaab - thanks for the contribution :) I'll need to find some time to read through this and the relevant links

mikefarah:
@vaab ok finally got around to reading the jq discussion and looking over the code. I think it does make sense to add this to `yq` 👍🏼 . The go-lang code looks pretty good to me, obviously missing unit tests. The acceptance tests ... it's clear to me that your bash skills are better than mine :D. To that end, I'd appreciate more comments in those tests so that I know what's going on :) I'll put in comments in the PR so you can see where.

Looking forward to getting this in!

acceptance_tests/nul-separator.sh
```diff
@@ -0,0 +1,251 @@
...
+## can't use NUL characters.
+assertFileSame() {
+    local fileA="$1" fileB="$2"
+
+    assertEquals "$(cat "$fileA" | hd)" \
+                 "$(cat "$fileB" | hd)"
+}
+
+
+read-0() {
```
mikefarah, line 9: I'm not sure what this does

acceptance_tests/nul-separator.sh
```diff
@@ -0,0 +1,251 @@
...
+    local eof="" IFS=''
+    while [ "$1" ]; do
+        read -r -d '' -- "$1" || eof=1
+        shift
+    done
+    [ -z "$eof" ]
+}
+
+
+read-0-err() {
```
mikefarah, line 29: I'm not sure what this does

acceptance_tests/nul-separator.sh
```diff
@@ -0,0 +1,251 @@
...
+    [ -z "$eof" ] || {
+        if [ "$last" != 0 ]; then
+            echo "Error: read-0-err couldn't fill all value" >&2
+            read -r -- "${ret?}" <<<"127"
+        fi
+        false
+    }
+}
+
+wyq() {
```
mikefarah: whats the `w` stand for here?

acceptance_tests/nul-separator.sh
```diff
@@ -0,0 +1,251 @@
...
+  boom
+EOL
+
+  read-0 a b c < <(./yq e -0 '.a, .b, .c' test.yml)
+  errlvl="$?"
+  assertNotEquals "0" "$errlvl"
+
+  ## using -r=false solves the issue but keeps all in yaml
+
+  read-0 a b c < <(./yq e -0 -r=false '.a, .b, .c' test.yml)
```
mikefarah, line 174: prefer to split this into another test

acceptance_tests/nul-separator.sh
```diff
@@ -0,0 +1,251 @@
...
+    cat >test.yml <<EOL
+- yay
+- wiz
+- "foo\0bar"
+- hop
+- pow
+EOL
+
+    res=""
+    while read-0 a || ! ret="$a"; do
```
mikefarah: this lines are black magic to me :)

acceptance_tests/nul-separator.sh
```diff
@@ -0,0 +1,251 @@
...
+    cat >test.yml <<EOL
+- yay
+- wiz
+- "foo\0bar"
+- hop
+- pow
+EOL
+
+    res=""
+    while read-0-err E a b; do
```
mikefarah, line 220: I have no idea what this does

vaab:
Many thanks for your interest, and your feedback make all perfect sense to me. I'll correct all this in the upcoming days and add some unit tests.

mikefarah:
ping!

vaab:
So sorry, had several things on my plate to deal with and I was not this good at organizing my time. I hope to finish this this week. I'm working on it right now. Thanks for your kind reminder.

vaab:
The code was rebased, your comments were taken into account (I have heavily commented and changed for the better some code on the bash part), and added some unit test as promised. 

Tell me your feelings. I'm sorry that it might require you to step a little in the murky waters of `bash`.

Thank you very much for your review and sorry for the delay.

mikefarah:
Thanks @vaab - very well commented bash :)

dezza:
Can you give an example of which elements are split with this? I tried today I could only get a nulbyte before the first element.

I also tried adding `---` between element prints and trying to make only a root element couldn't get it working.

vaab:
`yq`, depending on the query, will output 1 or more result. Every result that your query produce is separated with NUL char if using `-0` command line argument, otherwise, it's a newline char that is used.

You need to precisely master when `yq` produce one or more result. For instance:
```
$ echo "{a: b, c: d}" | yq 'keys()' | hd
00000000  2d 20 61 0a 2d 20 63 0a                           |- a.- c.|
00000008
```
The `keys()` returns a single YAML sequence `[a, c]` as a result, written in YAML:
```
- a
- c
```
This is exactly the content that you see in the hexdump output.
However, please notice that:
- The first `0a` in the hexdump is the newline between these 2 lines. That character  is part of YAML specification and it is the normal way to write valid YAML. 
- The second and final `0a` closes your first and only result `yq` result. This one is NOT required by YAML, and is added by `yq` to separate this result with possibly other results. This is the one that will be modified by `-0`:
```
$ echo "{a: b, c: d}" | yq 'keys()' -0 | hd
00000000  2d 20 61 0a 2d 20 63 00                           |- a.- c.|
00000008
```
To break down a YAML sequence to results, you can use `.[]`:
```
$ echo "{a: b, c: d}" | yq 'keys() | .[]' -0 | hd
00000000  61 00 63 00                                       |a.c.|
00000004
```
Here, `yq` outputs 2 results, spanning the different values that were in the YAML sequence `[a, b]`. As they are separate result, they are separated by NUL char. Without `-0`, these would be separated by `0a`... which would make the whole output of `yq` possibly ambiguous and difficult to parse depending on your situation.

Using `-0` as command line argument ensures you unambiguous manipulation of data, because NUL chars can't happen in valid YAML, 

Does that help ? You can share your examples if there's something still unclear in your situation.

dezza:
@vaab thanks a lot for that detailed breakdown, it helped me simplify because I was doing some `split_doc` madness and didn't know wtf I was doing or how to get it haha.. I just had to go down to a value structure basically without `key: value` sets..

PR #1582, opened 2023-03-02
Detect input format based on file name extension

### LINKED ISSUE #1440: Detect input format based on file name extension

Is it possible for yq to be a little more smart to detect the input file type based on the file extension? At least for common ones that yq already/would support, like:
- yml, yaml
- json
- xml
- csv
- properties

When an xml file (`*.xml`) is used as input, it should be obvious for yq to treat the input as xml, rather than yaml.

```xml
<p>
  <a>1</a> <!-- comment: message -->
</p>
```

And we run a command:
```bash
yq '... comments="" | .p' data1.xml
```

it could output
```yaml
a: "1"
```

it works by always explicitly specify the input file type, e.g.:
```bash
yq -px '... comments="" | .p' data1.xml
# a: "1"
```

**Additional context**
The request is related to https://github.com/mikefarah/yq/issues/1438#issuecomment-1319942685

ryenus:
When the inputFormat is not specified, instead of just assume the input format to be "yaml", we can also look at the input filename, more specifically the file extension, to deduce the input format. 

This Fixes #1440.

### DIFF
--- a/acceptance_tests/inputs-format-auto.sh
+++ b/acceptance_tests/inputs-format-auto.sh
@@ -1,0 +1,221 @@
+#!/bin/bash
+
+setUp() {
+  rm test*.yml 2>/dev/null || true
+  rm test*.json 2>/dev/null || true
+  rm test*.properties 2>/dev/null || true
+  rm test*.csv 2>/dev/null || true
+  rm test*.tsv 2>/dev/null || true
+  rm test*.xml 2>/dev/null || true
+}
+
+testInputJson() {
+  cat >test.json <<EOL
+{ "mike" : { "things": "cool" } }
+EOL
+
+  read -r -d '' expected << EOM
+mike:
+  things: cool
+EOM
+
+  X=$(./yq test.json)
+  assertEquals "$expected" "$X"
+
+  X=$(./yq ea test.json)
+  assertEquals "$expected" "$X"
+}
+
+testInputProperties() {
+  cat >test.properties <<EOL
+mike.things = hello
+EOL
+
+  read -r -d '' expected << EOM
+mike:
+  things: hello
+EOM
+
+  X=$(./yq e test.properties)
+  assertEquals "$expected" "$X"
+
+  X=$(./yq ea test.properties)
+  assertEquals "$expected" "$X"
+}
+
+testInputPropertiesGitHubAction() {
+  cat >test.properties <<EOL
+mike.things = hello
+EOL
+
+  read -r -d '' expected << EOM
+mike:
+  things: hello
+EOM
+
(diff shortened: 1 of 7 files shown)


mikefarah:
Looks really good - could you add some acceptance tests (see acceptance_tests folder) to test that the detection works for the various file types?

ryenus:
@mikefarah Thank you! I added a test, actually just copied but removing the explicit `-p` args.
Hopefully it works as expected. Just a humble attempt as a rare-time gopher, Please feel free to amend :-)

acceptance_tests/inputs-format-auto.sh
```diff
@@ -0,0 +1,221 @@
...
+EOL
+
+  read -r -d '' expected << EOM
+- fruit: apple
+  yumLevel: 5
+- fruit: banana
+  yumLevel: 4
+EOM
+
+  X=$(./yq e -p=csv test.csv)
```
mikefarah: still some `-p` to remove

mikefarah: Also - you need to make the script file executable (see build failure).

ryenus: Indeed, both fixed.

ryenus:
@mikefarah, any idea why `testLeadingSeparatorDoesNotBreakCommentsOnOtherFiles` is broken? Could you please take a look?

```
shunit2:ERROR testLeadingSeparatorDoesNotBreakCommentsOnOtherFiles() returned non-zero return code.
ASSERT:expected:<# a1
a: 1
# a2

# b1
b: 2
# b2> but was:<# a1
a: 1
# a2

b: 2
# b2>
```

mikefarah:
Looks like the `#b1` comment disappeared 🤔  not sure why that would happen

cmd/evaluate_all_command.go
```diff
@@ -75,7 +75,12 @@ func evaluateAll(cmd *cobra.Command, args []string) (cmdError error) {
 		return err
 	}
 
-	decoder, err := configureDecoder(true)
+	inputFilename := ""
+	if len(args) > 0 {
+		inputFilename = args[0]
+	}
+
+	decoder, err := configureDecoder(false, inputFilename)
```
mikefarah: found this issue - this has been set to false!

ryenus: Ahh, clearly a copy-paste mistake, fixed and pushed 😂

ryenus:
@hubchub, no, detecting **input** format based on the filename is not really breaking, as it works only when the **input** format is not specified. Meanwhile the explicitly set input format is always respected, just like how it behaved in the past. If you meant the change related to the default _output_ format, that's a separate topic.

Issue #1586, opened 2023-03-03
Inconsistency between `map` and `select`; proposal for a new `filter` function


rbren:
Apologies if this is a dupe, but I didn't see anything.

**Please describe your feature request.**
I'd like to use `map` and `select` the same way I use `map` and `filter` in JavaScript--to turn arrays into arrays. `map` does this very well already.

Consider this example:
```bash
arr="
- amt: 1
- amt: 2
- amt: 3
"
echo "${arr}" | yq e '. | map(.amt + 1)' -
echo "---"
echo "${arr}" | yq e '.[] | select(.amt > 1)' -
```

This outputs:
```yaml
- 2
- 3
- 4
---
2
3
```
Note that `map` outputs a valid YAML array, while `select` outputs raw lines of text.

**Describe the solution you'd like**
```bash
arr="
- amt: 1
- amt: 2
- amt: 3
"
echo "${arr}" | yq e '. | filter(.amt > 1)' -
```
which would output
```bash
- 2
- 3
```

Note that this makes chaining really nice! E.g.
```bash
echo "${arr}" | yq e '. | filter(.amt > 1) | map(.amt * 2) | filter(. > 3)' -
```

**Describe alternatives you've considered**
It's possible this has been solved by another function I'm not seeing.

**Additional context**

rbren:
Also--I'm happy to implement this if you think it's a good idea!

mikefarah:
Yeah nice - I think that would work well. I'd be very happy to take a PR :)

mikefarah:
Released in 4.32.1 - Thanks again @rbren  :)

PR #1588, opened 2023-03-06
Add filter operation

rbren:
Fixes https://github.com/mikefarah/yq/issues/1586


pkg/yqlib/operator_filter.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func filterOperator(d *dataTreeNavigator, context Context, expressionNode *ExpressionNode) (Context, error) {
+	log.Debugf("-- filterOperation")
+	var results = list.New()
+
+	for el := context.MatchingNodes.Front(); el != nil; el = el.Next() {
+		candidate := el.Value.(*CandidateNode)
+		children := context.SingleChildContext(candidate)
+		splatted, err := splat(children, traversePreferences{})
```
rbren, line 14: TBH, not sure why I'm using `splat`. Lots of copy/paste in here

pkg/yqlib/operator_filter.go
```diff
@@ -0,0 +1,56 @@
...
+			include, err := d.GetMatchingNodes(childCtx, expressionNode.RHS)
+			if err != nil {
+				return Context{}, err
+			}
+			var includeResult bool
+			var errDecoding error
+			includeEl := include.MatchingNodes.Front()
+			if includeEl.Next() != nil {
+				return Context{}, errors.New("Only expected one child")
+			}
```
rbren: This was the best way I could find to turn each value into a truthy value to determine if it should be included. LMK if there's a better way

rbren:
One thing I'm not super happy with is arrays-of-arrays. But it works the same for both `map` and `filter`.

```
❯ cat arr.yaml
- - 1
  - 2
  - 3
- - 4
  - 5
  - 6

❯ go run yq.go e '.[] | filter(. > 1)' arr.yaml
- 2
- 3
- 4
- 5
- 6

❯ go run yq.go e '.[] | map(. + 1)' arr.yaml
- 2
- 3
- 4
- 5
- 6
- 7
```

mikefarah:
Yep - the first expression splats the array of arrays, which becomes the new context- and then it maps (or filters) that context.

This is like the difference between: `.a | .b = 3` (which will set the root context to `.a` before updating `.b`, and doing `.a.b = 3` which will deeply update `b` without changing the root context.

To get what you want - you'd need to run the expression without changing the context. You can use 'with' to that:

```
yq 'with(.[]; map(. += 1))'
```

Hope that makes sense!

pkg/yqlib/operator_filter.go
```diff
@@ -0,0 +1,56 @@
...
+
+	for el := context.MatchingNodes.Front(); el != nil; el = el.Next() {
+		candidate := el.Value.(*CandidateNode)
+		children := context.SingleChildContext(candidate)
+		splatted, err := splat(children, traversePreferences{})
+		if err != nil {
+			return Context{}, err
+		}
+
+		if err != nil {
```
mikefarah, line 19: duplicate err check

pkg/yqlib/operator_filter.go
mikefarah: hmm, yeah there should be a cleaner way. What this function is basically doing is a map(select(X)). 

You could create a new `select` expression node and give the expression.RHS to it.

something like:

```
    selectNodesOp := &ExpressionNode{
      Operation: selectOp,
      RHS:       expression.RHS,
      }
      
      includedNodes, err := d.GetMatchingNodes(splatted, selectNodesOp.RHS)
```

Pretty sure this could be extended to create a map(select) expression and just delegate the logic to those functions entirely...

pkg/yqlib/operator_filter_test.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+var filterOperatorScenarios = []expressionScenario{
+	{
+		description: "Filter array",
+		document:   `[1,2,3]`,
+		expression: `filter(. < 3)`,
+		expected: []string{
+			"D0, P[], (!!seq)::- 1\n",
+			"D0, P[], (!!seq)::- 2\n",
```
mikefarah: this isn't actually the right result - this means the filter function is return two arrays - each with a single element. It should return a single array with two elements

```
"D0, P[], (!!seq)::- 1\n- 2\n",
```

I think if you refactor to use map and select under the hood - it would do the right thing.

pkg/yqlib/operator_filter.go
```diff
@@ -0,0 +1,38 @@
...
+
+	for el := context.MatchingNodes.Front(); el != nil; el = el.Next() {
+		candidate := el.Value.(*CandidateNode)
+		log.Warningf("candidate %#v", candidate)
+		children := context.SingleChildContext(candidate)
+		splatted, err := splat(children, traversePreferences{})
+		if err != nil {
+			return Context{}, err
+		}
+		filtered, err := selectOperator(d, splatted, expressionNode)
```
rbren, line 18: OK I think I have this in a better spot now. I'm pushing out most of the logic to `select`, and it seems like its generating the proper arrays

rbren:
Awesome, I didn't know about `with`! Interesting that it needs `+=` instead of just `+`

I also got it to do what I want with this expression:
```
go run yq.go e '.[] |= map(. + 1) | .[] |= filter(. < 4)' arr.yaml
```

fearphage:
Isn't `filter(...)` just shorthand for `map(select(...))`?

PR #1593, opened 2023-03-09
Added divide and modulo operators

### LINKED ISSUE #49: Add support for Operator '*, /, and %'
Add support for [Operator '*, /, and %'](https://stedolan.github.io/jq/manual/#example15)

```
These infix operators behave as expected when given two numbers. Division by zero raises an error. x % y computes x modulo y.

Multiplying a string by a number produces the concatenation of that string that many times. "x" * 0 produces null.

Dividing a string by another splits the first using the second as separators.

Multiplying two objects will merge them recursively: this works like addition but if both objects contain a value for the same key, and the values are objects, the two are merged with the same strategy.
```

teejaded:
fixes #49

### DIFF
--- a/pkg/yqlib/operator_divide.go
+++ b/pkg/yqlib/operator_divide.go
@@ -8,12 +8,6 @@ import (
 	yaml "gopkg.in/yaml.v3"
 )
 
-func createDivideOp(lhs *ExpressionNode, rhs *ExpressionNode) *ExpressionNode {
-	return &ExpressionNode{Operation: &Operation{OperationType: divideOpType},
-		LHS: lhs,
-		RHS: rhs}
-}
-
 func divideOperator(d *dataTreeNavigator, context Context, expressionNode *ExpressionNode) (Context, error) {
 	log.Debugf("Divide operator")
 
@@ -33,7 +27,7 @@ func divide(d *dataTreeNavigator, context Context, lhs *CandidateNode, rhs *Cand
 	target := &yaml.Node{}
 
 	if lhsNode.Kind == yaml.ScalarNode && rhs.Node.Kind == yaml.ScalarNode {
-		if err := divideScalars(context, target, lhsNode, rhs.Node); err != nil {
+		if err := divideScalars(target, lhsNode, rhs.Node); err != nil {
 			return nil, err
 		}
 	} else {
@@ -43,7 +37,7 @@ func divide(d *dataTreeNavigator, context Context, lhs *CandidateNode, rhs *Cand
 	return lhs.CreateReplacement(target), nil
 }
 
-func divideScalars(context Context, target *yaml.Node, lhs *yaml.Node, rhs *yaml.Node) error {
+func divideScalars(target *yaml.Node, lhs *yaml.Node, rhs *yaml.Node) error {
 	lhsTag := lhs.Tag
 	rhsTag := guessTagFromCustomType(rhs)
 	lhsIsCustom := false
--- a/pkg/yqlib/operator_divide.go
+++ b/pkg/yqlib/operator_divide.go
@@ -38,21 +38,14 @@ func divide(d *dataTreeNavigator, context Context, lhs *CandidateNode, rhs *Cand
 		Anchor: lhs.Node.Anchor,
 	})
 
-	switch lhsNode.Kind {
-	case yaml.MappingNode:
-		return nil, fmt.Errorf("%v (%v) cannot be divided by %v (%v)", lhsNode.Tag, lhs.GetNicePath(), rhs.Node.Tag, rhs.GetNicePath())
-
-	case yaml.SequenceNode:
-		return nil, fmt.Errorf("%v (%v) cannot be divided by %v (%v)", lhsNode.Tag, lhs.GetNicePath(), rhs.Node.Tag, rhs.GetNicePath())
-
-	case yaml.ScalarNode:
-		if rhs.Node.Kind != yaml.ScalarNode {
-			return nil, fmt.Errorf("%v (%v) cannot be divided by %v (%v)", lhsNode.Tag, lhs.GetNicePath(), rhs.Node.Tag, rhs.GetNicePath())
-		}
+	if lhsNode.Kind == yaml.ScalarNode && rhs.Node.Kind == yaml.ScalarNode {
 		if err := divideScalars(context, target, lhsNode, rhs.Node); err != nil {
 			return nil, err
 		}
+	} else {
+		return nil, fmt.Errorf("%v (%v) cannot be divided by %v (%v)", lhsNode.Tag, lhs.GetNicePath(), rhs.Node.Tag, rhs.GetNicePath())
 	}
+
 	return target, nil
 }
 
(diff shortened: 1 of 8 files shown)


pkg/yqlib/operator_divide.go
```diff
@@ -0,0 +1,88 @@
...
+
+		lhsNum, err := strconv.ParseFloat(lhs.Value, 64)
+		if err != nil {
+			return err
+		}
+		rhsNum, err := strconv.ParseFloat(rhs.Value, 64)
+		if err != nil {
+			return err
+		}
+		quotient := lhsNum / rhsNum
```
mikefarah, line 67: what happens if you divide by zero?

mikefarah: (be good to have a test so its in the docs)

teejaded:
```
./yq -n '1 / 0'
+Inf
```

I'll add a test.

pkg/yqlib/operator_divide_test.go
```diff
@@ -0,0 +1,121 @@
...
+			"D0, P[], (doc)::{a: 4.8, b: 2.5}\n",
+		},
+	},
+	{
+		skipDoc:     true,
+		description: "Custom types: that are really strings",
+		document:    "a: !horse cat_meow\nb: !goat _",
+		expression:  `.a = .a / .b`,
+		expected: []string{
+			"D0, P[], (doc)::a: !horse\n    - cat\n    - meow\nb: !goat _\n",
```
mikefarah, line 57: in this case I think it'd be more correct if `a` is an array, but each element is a `horse`

pkg/yqlib/operator_modulo.go
```diff
@@ -0,0 +1,98 @@
...
+	rhs.Node = unwrapDoc(rhs.Node)
+
+	lhsNode := lhs.Node
+
+	if lhsNode.Tag == "!!null" {
+		return nil, fmt.Errorf("%v (%v) cannot modulo by %v (%v)", lhsNode.Tag, lhs.GetNicePath(), rhs.Node.Tag, rhs.GetNicePath())
+	}
+
+	target := lhs.CreateReplacement(&yaml.Node{
+		Anchor: lhs.Node.Anchor,
```
mikefarah: not sure why this is here? (maybe I need it in the sub/add/mult operators ?

pkg/yqlib/operator_modulo.go
```diff
@@ -0,0 +1,98 @@
...
+
+		format, lhsNum, err := parseInt64(lhs.Value)
+		if err != nil {
+			return err
+		}
+		_, rhsNum, err := parseInt64(rhs.Value)
+		if err != nil {
+			return err
+		}
+		remainder := lhsNum % rhsNum
```
mikefarah, line 66: again good to test what happens if rhs is 0

mikefarah:
Thanks @teejaded  - this looks promising :)

pkg/yqlib/operator_modulo.go
teejaded: Yeah I'm not sure.  I started with the add operator which is where it came from https://github.com/mikefarah/yq/blob/master/pkg/yqlib/operator_add.go#L59-L61

pkg/yqlib/operator_modulo.go
teejaded: https://github.com/mikefarah/yq/pull/1593/commits/c0fd3fe38b756eef3745db4019fce23a119bd6dc

pkg/yqlib/operator_divide.go
teejaded: https://github.com/mikefarah/yq/pull/1593/commits/16997769e0146cf285e9a75a5e7c27300fbb7ffb

pkg/yqlib/operator_divide_test.go
teejaded: The behavior came from string split.  Should I try to fix it there?

```bash
❯ echo $'a: !horse cat_meow\nb: !goat _' | yq '.a | split("_")'
- cat
- meow

❯ echo $'a: !horse cat_meow\nb: !goat _' | yq '.a = (.a |split("_"))'
a: !horse
  - cat
  - meow
b: !goat _
```

pkg/yqlib/operator_modulo.go
teejaded: I think it isn't necessary because I didn't make `/=` or `%=` operators.  I'll remove it.

teejaded: https://github.com/mikefarah/yq/pull/1593/commits/b499322914d56e49af2e577be9cc9ff9c43dad1c

pkg/yqlib/operator_divide_test.go
mikefarah: oh I see - yeah I think so 🤔 , that would be appreciated

pkg/yqlib/operator_divide.go
teejaded: I was wondering this morning if +Inf is supported in json, and I guess it is not. NaN is the same.  What's your opinion on this?

https://stackoverflow.com/questions/1423081/json-left-out-infinity-and-nan-json-status-in-ecmascript



```bash
./yq -o json -n '{"a": 1 / 0}'
Error: yaml: cannot decode !!str `+Inf` as a !!float

jq -n '{"a": +Inf}'
jq: error: syntax error, unexpected '+' (Unix shell quoting issues?) at <top-level>, line 1:
{"a": +Inf}
jq: 1 compile error

 jq -n '{"a": infinite}'
{
  "a": 1.7976931348623157e+308
}
```

https://stedolan.github.io/jq/manual/

>  infinite, nan, isinfinite, isnan, isfinite, isnormal
> 
> Some arithmetic operations can yield infinities and "not a number" (NaN) values. The isinfinite builtin returns true if its input is infinite. The isnan builtin returns true if its input is a NaN. The infinite builtin returns a positive infinite value. The nan builtin returns a NaN. The isnormal builtin returns true if its input is a normal number.
> 
> Note that division by zero raises an error.
> 
> Currently most arithmetic operations operating on infinities, NaNs, and sub-normals do not raise errors.

pkg/yqlib/operator_divide_test.go
teejaded: Actually, I'm not sure I can change it easily.  It seems the assignment operator won't change custom tags.  Changing that would be a breaking change.

```bash
❯ echo $'a: !horse cat_meow\nb: !goat _' | yq '.a = {} | .c = {} | .[] | tag'
!horse
!goat
!!map

❯ echo $'a: !horse cat_meow\nb: !goat _' | yq '.a = {} | .c = {} | .c = [] | .[] | tag'
!horse
!goat
!!seq
```

mikefarah: alright - leave it, no worries

mikefarah:
Couple of failures in the pipeline - should be easy to fix. Just run `make local build` to reproduce

teejaded:
I think it's fixed now!  Sorry I neglected that step from CONTRIBUTING.md.

PR #1604, opened 2023-03-16
Add support for decoding base64 strings without padding

### LINKED ISSUE #1555: @base64d fails with base64 encoded string without padding
**Describe the bug**
I am trying to use yq to decode the payload of a JWT token.

While this works with jq using
```sh
$ jq -R 'split(".") | .[1] | @base64d | fromjson' <<< "$MY_TOKEN"
{
  //...
}
```
It fails with yq:
```sh
$ yq 'split(".") | .[1] | @base64d | fromjson' <<< "$MY_TOKEN"
Error: unexpected EOF
```

yq fails because JWT tokens consist of base64 components that are not padded using equals signs.

```sh
$ echo "dGVzdA==" | yq "@base64d"
test
$ echo "dGVzdA" | yq "@base64d"
Error: unexpected EOF
```

Version of yq: 4.30.8
Operating systems: mac/linux/windows
Installed via: binary release

**Actual behavior**

```sh
$ echo "dGVzdA" | yq "@base64d"
Error: unexpected EOF
```

**Expected behavior**

yq can handle base64 strings that are not padded.

```sh
$ echo "dGVzdA" | yq "@base64d"
test # this should be returned
```

**Additional context**

teejaded:
fixes #1555

Kinda tricky to fix because the encode decode operator is very generic so it's an awkward place to put a conditional to handle just unpadded base64 strings.

The base64Decoder itself just holds a Reader so it had no idea how long the data is.

### DIFF
--- a/pkg/yqlib/decoder_base64.go
+++ b/pkg/yqlib/decoder_base64.go
@@ -4,9 +4,32 @@ import (
 	"bytes"
 	"encoding/base64"
 	"io"
+	"strings"
 
 	yaml "gopkg.in/yaml.v3"
 )
+
+type base64Padder struct {
+	count uint64
+	io.Reader
+}
+
+func (c *base64Padder) pad(buf []byte) (int, error) {
+	pad := strings.Repeat("=", int(4-c.count%4))
+	n, err := strings.NewReader(pad).Read(buf)
+	c.count += uint64(n)
+	return n, err
+}
+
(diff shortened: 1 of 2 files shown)


teejaded:
I also couldn't find a way to configure the library to accept padded and not padded data.

https://go.dev/play/p/v8NJHgxKdZx

mikefarah:
Nice - I like the counting approach

Issue #1632, opened 2023-04-06
Output to shell variables


giorgiga:
It would be helpful if yq could output values as VARIABLE=VALUE lines to be sourced in bash.

The format would be similar to the existing props one, but keys should be sanitized (no dots, no hyphens ...) and values should be properly escaped.

A variant that outputs fish `set` commands instead of posix VARIABLE=VALUE would also be great, but there are utilities that convert posix to fish.

Suppose we have:

```yaml
country: 
  name: Australia
  capital: Camberra
```

And we run a command:

```bash
yq -o shvars
```

it could output

```yaml
COUTNRY_NAME='Australia'
COUNTRY_CAPITAL='Camberra'
```

mikefarah:
Ill accept a PR for this - in the meantime you can get a similar result by:

```
yq '.. | select(tag == "!!str") |  (path | join("_") | upcase) + " = " + .' examples/data1.yaml
```

PR #1645, opened 2023-04-23
Implement shell output format

giorgiga:
This implements the idea at #1632 to export in a format that can be sourced into bash/sh.

Since both yaml keys and shell variables are case-sensitive, it seems converting to uppercase as I imagined while writing #1632 is not the best of ideas after all :sweat_smile:.

If case is preserved, there are instances where the yaml keys will not correspond to the exported variable (because variable names are quite restrictive compared to yaml). see the comments in `encoder_shellvariables.go` for details.

Eg. `echo -e '1:\n "a a": somevalue' | go run yq.go -o shell` will output `_1_a_a=somevalue` (note the initial underscore and the underscore that replaces the space between the to 'a').

Documents  consisting of a single scalar value are converted as if a key named "value" was present. This is because it does not make sense to have variables with empty names (and `_` is  a special shell variable that can't be assigned to).

Eg. `echo somevalue | go run yq.go -o shell` will print `value=somevalue`.

I've not updated the documentation yet.

### DIFF
--- a/pkg/yqlib/encoder_shellvariables.go
+++ b/pkg/yqlib/encoder_shellvariables.go
@@ -1,0 +1,153 @@
+package yqlib
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"unicode/utf8"
+
+	"golang.org/x/text/unicode/norm"
+	yaml "gopkg.in/yaml.v3"
+)
+
+type shellVariablesEncoder struct {
+}
+
+func NewShellVariablesEncoder() Encoder {
+	return &shellVariablesEncoder{}
+}
+
+func (pe *shellVariablesEncoder) CanHandleAliases() bool {
+	return false
+}
+
+func (pe *shellVariablesEncoder) PrintDocumentSeparator(_ io.Writer) error {
+	return nil
+}
+
+func (pe *shellVariablesEncoder) PrintLeadingContent(_ io.Writer, _ string) error {
+	return nil
+}
+
+func (pe *shellVariablesEncoder) Encode(writer io.Writer, node *yaml.Node) error {
+
+	mapKeysToStrings(node)
+	err := pe.doEncode(&writer, node, "", nil)
+	if err != nil {
+		return err
+	}
+
+	return err
+}
+
+func (pe *shellVariablesEncoder) doEncode(w *io.Writer, node *yaml.Node, path string, _ *yaml.Node) error {
+
+	// Note this drops all comments.
+
+	switch node.Kind {
+	case yaml.ScalarNode:
+		nonemptyPath := path
+		if path == "" {
+			// We can't assign an empty variable "=somevalue" because that would error out if sourced in a shell,
+			// nor can we use "_" as a variable name ($_ is a special shell variable that can't be assigned)...
+			// let's just pick a fallback key to use if we are encoding a single scalar
+			nonemptyPath = "value"
+		}
+		_, err := io.WriteString(*w, nonemptyPath+"="+quoteValue(node.Value)+"\n")
+		return err
+	case yaml.DocumentNode:
+		return pe.doEncode(w, node.Content[0], path, node)
+	case yaml.SequenceNode:
+		for index, child := range node.Content {
+			err := pe.doEncode(w, child, appendPath(path, index), nil)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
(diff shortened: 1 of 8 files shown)


mikefarah:
Looks good! For testing (and documentation) - can you follow the pattern in TestPropertyScenarios (albeit you'll only have encode tests). That will also generate documentation

giorgiga:
Forgot to mention: the makefile doesn't work for me...

I get (for example):
````
$  make vendor
/usr/bin/podman run --rm -e LDFLAGS="-X main.GitCommit=6017174 -X main.GitDescribe=v4.33.3-6-g6017174 -w -s" -e GITHUB_TOKEN="" -v /home/giorgio/projects/yq/vendor:/go/src -v /home/giorgio/projects/yq:/yq/src/github.com/mikefarah/yq -w /yq/src/github.com/mikefarah/yq yq_dev go mod vendor
✔ docker.io/library/yq_dev:latest
Trying to pull docker.io/library/yq_dev:latest...
Error: initializing source docker://yq_dev:l
````

I run the `scripts/` directly without a container so I guess it doesn't matter for this PR, but I thought you might have wanted to know (as for what the problem is and if it's  the makefile or my machine, I have no idea)

**Comment by mikefarah** [comment]
Hmm - I run everything locally (make local build) rather than docker/podman.

Markdown is all that's needed :) The service I use for documentation does the conversion to HTML itself.

File: pkg/yqlib/encoder_shellvariables.go [hunk 1]
```diff
@@ -0,0 +1,153 @@
...
+	mapKeysToStrings(node)
+	err := pe.doEncode(&writer, node, "", nil)
+	if err != nil {
+		return err
+	}
+
+	return err
+}
+
+func (pe *shellVariablesEncoder) doEncode(w *io.Writer, node *yaml.Node, path string, _ *yaml.Node) error {
```
mikefarah: Not sure why you have the last argument here if it's just ignored?

pkg/yqlib/encoder_shellvariables.go
```diff
@@ -0,0 +1,153 @@
...
+	//     3. replace all non-alphanumeric characters with _
+	//
+	// Moreover, for the root key only, we will prepend an underscore if what results from the steps above
+	// does not start with [a-zA-Z_] (ie. if the root key starts with a digit).
+	//
+	// Note this is NOT a 1:1 mapping.
+	//
+	// (*) see endofname.c from https://git.busybox.net/busybox/tag/?h=1_36_0
+
+	// XXX empty strings
```
mikefarah, line 112: love this comment block - well explained thankyou :)

pkg/yqlib/encoder_shellvariables.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+	if needsQuoting {
+		return "'" + strings.ReplaceAll(value, "'", "'\"'\"'") + "'"
+	}
+	return value
+}
+
+func isAlphaOrUnderscore(r rune) bool {
+	return ('a' <= r && r <= 'z') || ('A' <= r && r <= 'Z')
```
mikefarah: doesn't this need `|| r == '_'` ?

pkg/yqlib/encoder_shellvariables.go
giorgiga: It's vestigial remains from the properties encoder that I initially cop-ypasted - good catch!

pkg/yqlib/encoder_shellvariables.go
giorgiga: The function was initially just `isAlpha`... I must have thought renaming was all it needed to change the implementation!

I'm adding a test case case for this... well, for when the root key starts with an underscore, which is why this function should allow `_`.

pkg/yqlib/encoder_shellvariables.go
giorgiga: Fixed with https://github.com/mikefarah/yq/pull/1645/commits/8902f98db16ada13a2404eadda5d38bb6dabc7cc

pkg/yqlib/encoder_shellvariables.go
giorgiga: Fixed with https://github.com/mikefarah/yq/pull/1645/commits/8902f98db16ada13a2404eadda5d38bb6dabc7cc

PR #1734, opened 2023-07-20
document how to add anchors/alias to array

peterromfeldhk:
i also added the alias but its a bit hacky, please advise if there is a better way


mikefarah:
Hey, thanks for the PR, yes there is a better way :) those files are actually generated from tests, for alias see: https://github.com/mikefarah/yq/blob/master/pkg/yqlib/operator_anchors_aliases_test.go

peterromfeldhk:
sorry i am having troubles reading and understanding these tests :(

PR #1739, opened 2023-07-25
Changes the sample version in the installation instructions

renatomjr:
The previous `v4.2.0` in the example is very outdated.


mikefarah:
That's provided as an example of how to specify the version - it's not intended to show the latest version

PR #1784, opened 2023-09-14
fix typo

imishinist:
I found another typo, should I address that one as well?


mikefarah:
Thanks, Ill run cspell to find the issues.

PR #1745, opened 2023-07-30
Implement basic Lua output support

Zash:
First attempt at #1700 

Supports basic serialization to Lua tables:

```
$ ./yq -n  -olua '.hello="\"world\"",.foo=["bar","baz",3,1.3]' 
return {["hello"]="\"world\"";["foo"]={"bar","baz",3,1.3,};};
```

No indentation or pretty printing yet, likely next steps.

Feedback appreciated, hence the Draft PR. (Haven't written Go in a couple of years so be gentle.)

Issue #1700, opened 2023-06-17
Lua table support


Zash:
**Please describe your feature request.**
A clear and concise description of what the request is and what it would solve. 
Eg. I wish I could use yq to support [Lua](https://www.lua.org/) [tables](https://www.lua.org/manual/5.4/manual.html#3.4.9) as input and output formats. This would allow querying and manipulating Lua based file formats like [LuaRocks](https://luarocks.org/) [rockspec format](https://github.com/luarocks/luarocks/wiki/Rockspec-format) or the [Prosody](https://prosody.im/) `internal` data storage format.

**Describe the solution you'd like**
If we have data1.yml like:
(please keep to around 10 lines )

```yaml
country: Australia # this place
cities:
- Sydney
- Melbourne
- Brisbane
- Perth
```

And we run a command:

```bash
yq --output-format lua data1.yml
```

it could output

```lua
country = "Australia" -- this place
cities = {
  "Sydney",
  "Melbourne",
  "Brisbane",
  "Perth",
}
```

or alternatively (perhaps a flag)

```lua
return {
  country = "Australia"; -- this place
  cities = {
    "Sydney",
    "Melbourne",
    "Brisbane",
    "Perth",
  }
}
```

Lua tables support arbitrary types as both keys and values, but in stored serialized formats this is rarely used, so keeping to a subset such as only allowing scalar keys may be sensible.

**Describe alternatives you've considered**
Writing simple converters to/from Lua and yaml/json etc is relatively easy, I have a few of these already, but they lack the query and manipulation capabilities and usually don't preserve comments or order of keys like `yq` does.

mikefarah:
Fixed in 4.35.1

Zash:
Input support would be cool too.

PR #1810, opened 2023-09-30
Add support for Lua input

Zash:
The missing input half of #1700 / #1745  

Uses a full blown Lua parser, it will happily `while true do --[[ nothing, forever ]] end` so best keep away from untrusted input. Be on the lookout for a way to limit this or a more limited parser. 😅 

Given a sample.lua file of:
```lua
return {
        ["country"] = "Australia"; -- this place
        ["cities"] = {
                "Sydney",
                "Melbourne",
                "Brisbane",
                "Perth",
        };
};
```
then
```bash
yq -oy '.' sample.lua
```
will output
```yaml
country: Australia
cities:
  - Sydney
  - Melbourne
  - Brisbane
  - Perth
```

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -130,6 +130,8 @@ func configureDecoder(evaluateTogether bool) (yqlib.Decoder, error) {
 
 func createDecoder(format yqlib.InputFormat, evaluateTogether bool) (yqlib.Decoder, error) {
 	switch format {
+	case yqlib.LuaInputFormat:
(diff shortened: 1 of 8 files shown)


mikefarah:
Brilliant

stellarpower:
Hi, I have lua tables as config files without the "return" from [xmake](https://xmake.io). I can echo return in a shell pipeline easily enough, but, it'd also be nice if yq had support for this :).
Would it be easy enough to add? Not really familiar with lua - so IDK how prolific this kind of config storage on-disc is with lua-based tools; feel I've seen it before though.

PR #1811, opened 2023-10-04
Lua output fixes

Zash:
Fixes for some quirks noticed while working on #1810 

- `+.inf` was not handled
- `yq -o lua .foo <<< '{"foo":{"bar":1}}` or the `--prettyPrint` flag would output `{bar=1}` only, without the pre- & suffix.
- Handle empty `Tag` field, if the parser side does not set it (I did before finding a good way to distinguish integer and float in the Lua parser)

Could split into multiple PRs if that is desirable.

### DIFF
--- a/acceptance_tests/output-format.sh
+++ b/acceptance_tests/output-format.sh
@@ -271,4 +271,43 @@ EOM
   assertEquals "$expected" "$X"
 }
 
+testLuaOutputPretty() {
+  cat >test.yml <<EOL
+animals:
+  cat: meow
+EOL
+
+  read -r -d '' expected << EOM
+return {
+	["animals"] = {
+		["cat"] = "meow";
+	};
+};
+EOM
+
+  X=$(./yq e --output-format=lua test.yml)
+  assertEquals "$expected" "$X"
+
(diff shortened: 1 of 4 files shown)


mikefarah:
Is it possible to add a test to show the prettyPrint fix? If it can't be done in lua_test.go you could always add it in as an acceptance tests shell file (under output_format.sh will probably do).

Zash:
Will try.

Zash:
Before

```
...
testLuaOutputPretty
ASSERT:expected:<return {
	["animals"] = {
		["cat"] = "meow";
	};
};> but was:<{
	["animals"] = {
		["cat"] = "meow";
	};
}>
shunit2:ERROR testLuaOutputPretty() returned non-zero return code.
testLuaOutputSubset
ASSERT:expected:<return {
	["cat"] = "meow";
};> but was:<{
	["cat"] = "meow";
}>
shunit2:ERROR testLuaOutputSubset() returned non-zero return code.
```

After

```
...
testLuaOutputPretty
testLuaOutputSubset

Ran 17 tests.

OK
```

PR #1814, opened 2023-10-05
Fixes #1692 - Inconsistent backslash escaping in literal strings

### LINKED ISSUE #1692: Backslash duplication in literal strings
**Describe the bug**
Can't write a literal string with a single backslash in it.  To obtain a string with a single backslash requires the use of other functions.

Version of yq: 4.33.3
Operating system: mac
Installed via: binary release

**Command**
The command you ran:
```
yq -nr eval '"\\"'
yq -nr eval '"\"'
yq -nr eval '"\\"|match("^(.)")|.string'
```

**Actual behavior**

```
\\
Error: 1:1: invalid input text "\"\\\""
\
```

**Expected behavior**

Either:
```
\
Error: 1:1: invalid input text "\"\\\""
\
```

Or:
```
\\
\
\
```

**Additional context**
It appears that the only backslash escape that is recognised is `\n`.  There is no `\r`, `\t`, `\b`, `\xHH` or `\uXXXX`.  All other alphabetics are left alone with the preceding backslash left in the string.

I also tried to use single quotes and backquotes in case they were alternate string literal expressions but they also just gave the `invalid input text` error.

The actual workaround I am currently using is:
```
BACKSLASH='\' yq -nr eval 'strenv(BACKSLASH)'
```

(It's actually a much more complicated expression that happens to include `... | split(strenv(BACKSLASH)) | ...` as part of it)

KrisShannon:
This is the simplest fix I can come up with for #1692

Before:
```sh
% yq -nr eval '"\\"'
\\
```

After:
```sh
% yq -nr eval '"\\"'
\
```

It does change behaviour.  Anyone who is currently using double backslashes in their literal strings will need to change them to quadruple backslashes.

It also makes it possible to use a literal string of a windows path for a filename that starts with `n`:
```
% yq -nr eval '"C:\\tmp\\nigel.xlsx"'
C:\tmp\nigel.xlsx
```

### DIFF
--- a/pkg/yqlib/lexer_participle.go
+++ b/pkg/yqlib/lexer_participle.go
@@ -369,8 +369,8 @@ func stringValue() yqAction {
 		log.Debug("rawTokenvalue: %v", rawToken.Value)
 		value := unwrap(rawToken.Value)
 		log.Debug("unwrapped: %v", value)
-		value = strings.ReplaceAll(value, "\\\"", "\"")
-		value = strings.ReplaceAll(value, "\\n", "\n")
+		replacer := strings.NewReplacer("\\\\", "\\", "\\\"", "\"", "\\n", "\n")
+		value = replacer.Replace(value)
 		log.Debug("replaced: %v", value)
 		return &token{TokenType: operationToken, Operation: createValueOperation(value, value)}, nil
 	}


KrisShannon:
It may make sense to extract the creation of the string Replacer outside of the `stringValue` function and use a single Replacer for all literals instead of a new one each time.  I'm not a golang programmer so I don't know what the best way to do that would be.

mikefarah:
Can you add some tests for this :)

mikefarah:
Closing this till unit tests are added

PR #1825, opened 2023-10-13
Missing dependencies to build application

smutel:
Currently it's not possible to build the application with `make` due to missing dependencies in the docker container.
The goal of this PR is to add these missing dependencies.


mikefarah:
Not sure what VCS stamping is

PR #1837, opened 2023-10-22
add tests for root node unwrap bug in previous versions

vaab:
Well, I found a bug, and was ready to fix it when I saw that your recent commit labelled `Generic AST` fixed it. If you are interested, here are the tests I wrote before trying to fix this bug. Every currently published releases (up to `v4.35.2`) have this bug:

```
$ echo "'a'" | ./yq -r '.'  ## fails to unwrap
'a'
$ echo "'a'" | ./yq -r '. + ""'  ## works fine
a
```

### DIFF
--- a/acceptance_tests/output-format.sh
+++ b/acceptance_tests/output-format.sh
@@ -72,6 +72,15 @@ EOL
   assertEquals "\"cat\"" "$X"
 }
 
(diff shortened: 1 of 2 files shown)


mikefarah:
Awesome - always happy to welcome more tests

PR #1914, opened 2024-01-07
snap: update and migrate to core22.

soumyaDghosh:
soon, core24 is gonna release. But for now core22 is stable and a better option. So, ported to core24.


mikefarah:
Thanks, going to apply and test this locally

PR #1936, opened 2024-02-04
fix: empty TOML table

### LINKED ISSUE #1924: Rejects empty TOML tables
**Describe the bug**

yq rejects  perfectly legal empty TOML tables 
Version of yq: v4.40.5

**Input Toml**

input.toml:
```
[emptyTable]
```

**Command**
```
yq -o=json '.' input.toml
```

**Actual behavior**
```
Error: bad file 'input.toml': error retrieving table [emptyTable] value: %!w(<nil>)
```

**Expected behavior**

Parsing works. https://toml.io/en/v1.0.0#table says: "Empty tables are allowed and simply have no key/value pairs within them."

elibroftw:
- Added empty table unit test

I moved the code so that the if hasValue fails, it first checks for the empty table. The tests all pass, so not sure what

Closes #1924

### DIFF
--- a/pkg/yqlib/decoder_toml.go
+++ b/pkg/yqlib/decoder_toml.go
@@ -267,6 +267,12 @@ func (dec *tomlDecoder) processTable(currentNode *toml.Node) (bool, error) {
 	fullPath := dec.getFullPath(currentNode.Child())
 	log.Debug("!!!fullpath: %v", fullPath)
 
+	tableValue := dec.parser.Expression()
+	if tableValue.Kind != toml.KeyValue {
+		log.Debug("got an empty table, returning")
+		return true, nil
+	}
+
 	hasValue := dec.parser.NextExpression()
 	if !hasValue {
 		return false, fmt.Errorf("error retrieving table %v value: %w", fullPath, dec.parser.Error())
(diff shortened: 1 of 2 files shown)


pkg/yqlib/decoder_toml.go
```diff
@@ -269,6 +269,11 @@ func (dec *tomlDecoder) processTable(currentNode *toml.Node) (bool, error) {
 
 	hasValue := dec.parser.NextExpression()
 	if !hasValue {
```
mikefarah, line 277: It seems like this whole block could move _after_ the check below where the code already checks for an empty table, then the code will have already exited if it was an empty table.

pkg/yqlib/toml_test.go
```diff
@@ -199,6 +205,12 @@ var tomlScenarios = []formatScenario{
 		expected:     sampleArrayTableExpected,
 		scenarioType: "decode",
 	},
+	{
+		description:  "Parse: Empty Table",
+		input:        emptyTable,
+		expected:     emptyTableExpected,
+		scenarioType: "decode",
```
mikefarah, line 212: 👍🏼  for writing a test :)

PR #1981, opened 2024-03-14
doc: minor typo fix on example

stelleal:
Fixes a typo found in the example in the following topic in the docs:
https://mikefarah.gitbook.io/yq/usage/tips-and-tricks#updating-deeply-selected-paths

While I was reading the documentation and testing this (amazing) tool, I came across an error in the example that this Pull Request fixes. My tests involved rewriting a given file with the flag -i.

As it stands, when trying to run the example command an error similar to this is returned:
`Error: 1:30: invalid input text "name == \"fred) | ..."`

The proposed fix brings the example into line with expected behavior.


mikefarah:
Fixed thanks

PR #1992, opened 2024-03-27
min/max operators

mbenson:


mbenson:
this could probably be done with compare and reduce, but I imagine the same could be said for several other operators.

mikefarah:
👍🏼 👍🏼

PR #2112, opened 2024-07-23
fix out of range panic

AdamKorcz:
I am getting a case where `index` is `0` and the line therefore attempts to access `tokens[-1]`.


mikefarah:
Thanks for the PR! Are you able to add a test as well? Probably lexer_participle_test would be a good spot...
Otherwise if you let me know what the scenario was, I can add one myself as well.

PR #2192, opened 2024-11-11
chore: update golintci-liner and fix linting issues 

mavimo:
Each commit is going to apply a change to make the linter error free

@mikefarah after this you can consider to make the check mandatory in CI

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,7 +10,9 @@ linters:
     - gofmt
     - goimports
     - gosec
-    - megacheck
+    - gosimple
+    - staticcheck
+    - unused
     - misspell
     - nakedret
     - nolintlint
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -34,12 +34,5 @@ linters-settings:
 issues:
   exclude-rules:
   - linters:
-    - gosec
-    text:  "Implicit memory aliasing in for loop."
-    path: _test\.go
-  - linters:
-    - revive
-    text: "unexported-return"
-  - linters:
     - revive
     text: "var-naming"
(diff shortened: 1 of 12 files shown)


.golangci.yml
```diff
@@ -10,7 +10,9 @@ linters:
     - gofmt
     - goimports
     - gosec
-    - megacheck
+    - gosimple
+    - staticcheck
+    - unused
     - misspell
     - nakedret
     - nolintlint
```
ccoVeille, line 18: I would like to suggest to configure nolintlint stricter 

https://golangci-lint.run/usage/linters/#nolintlint

Having `nolint: revive` added without explanation is a bit a problem because then you have no idea why it was added. revive linters has so many rules...

Of course, this could be part of another PR.

I have no idea about the impact of such changes on the existing nolint directives present in the code.

pkg/yqlib/decoder_toml.go
```diff
@@ -286,7 +286,7 @@ func (dec *tomlDecoder) processTable(currentNode *toml.Node) (bool, error) {
 		}
 
 		runAgainstCurrentExp, err = dec.decodeKeyValuesIntoMap(tableNodeValue, tableValue)
-		if err != nil && !errors.Is(io.EOF, err) {
+		if err != nil && !errors.Is(err, io.EOF) {
```
ccoVeille, line 289: This one introduce a real bug fix and a change in the code behavior maybe.

Unless the error was not wrapped at all

.golangci.yml
mavimo: @ccoVeille thx for the feedback, I'll check how to make it stricter and I'll add a comment to explain the reason (BTW: we can avoid it if we replace the var name with `_`, but I prefer to keep the same name used in cobra)

mikefarah: yeah I don't understand the point of `//nolint:revive`?

pkg/yqlib/decoder_toml.go
mikefarah: While that's more correct (now that I read the docs again) - afaik this doesn't fix a known bug as the err is an instance of io.EOF anyway

.golangci.yml
mikefarah: Ok ran the build - I see. I think I'd prefer to rename the variables and not have an exception 👍🏼

PR #2231, opened 2024-12-13
feat: Create parent directories if --split-exp is used.

rudo-thomas:
Problem: When --split-exp is used and produces filenames with slashes in them, the target directories must already exist otherwise yq fails.

Fix/feature: Create the necessary directories with os.MkdirAll(). The permissions 0750 were chosen to satisfy the vulnerability checker.

### DIFF
--- a/acceptance_tests/split-printer.sh
+++ b/acceptance_tests/split-printer.sh
@@ -2,6 +2,7 @@
 
 setUp() {
   rm test*.yml || true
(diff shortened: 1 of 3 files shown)


rudo-thomas:
FWIW, the `jsonnet` binary has an extra flag for enabling the creation of the directories in a similar case.

I didn't think it was necessary here...

PR #2257, opened 2025-01-13
update ubuntu packages

angeling11:


angeling11:
update ubuntu dependencies to patch vulnerabilities

PR #2334, opened 2025-04-02
Add s390x platform support to Docker release workflow

ashokpariya0:
Updated the list of supported platforms in the GitHub Actions Docker release workflow to include linux/s390x.


ashokpariya0:
cc @mikefarah

PR #2344, opened 2025-04-08
Enable builds for Windows ARM

ShukantPal:
Hi @mikefarah, great work on yq! Right now, yq runs in compatibility mode on Windows but it would be nice to have a native Windows ARM build. This also makes it easier to make Bazel builds work where yq is a dependency.


mikefarah:
Thanks for the PR - there's a conflict, I'll just make the one liner separately 👍🏼

Issue #2348, opened 2025-04-23
add `arm64` binary for Windows


albertocavalcante:
Currently `yq` has both `windows_amd64` and `windows_386` but no `windows_arm64`. Can it be included, please? Thanks!

Related to: https://github.com/mikefarah/yq/issues/168

PR #2349, opened 2025-04-23
add `windows_arm64` binary

albertocavalcante:
Fixes https://github.com/mikefarah/yq/issues/2348

I've also made some changes to `scripts/` since the current config for `golangci-lint` fails with >=v2, and replaced `wget` with `curl` for consistency and to work better with `macOS`.


albertocavalcante:
@mikefarah can you please review when you get a moment? Thanks!

PR #2373, opened 2025-05-15
Marshal goccy

LarsStegman:
We have a project in which we parse a YAML file to a Goccy AST, then use yq (embedded in the yaml file) to transform the AST. The step we are missing is to be able to marshal from the yq CandidateNode back to a the Goccy AST. If there is a better way to do this, I am open to it :)

### DIFF
--- a/pkg/yqlib/candidate_node_goccy_yaml.go
+++ b/pkg/yqlib/candidate_node_goccy_yaml.go
@@ -2,6 +2,7 @@ package yqlib
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	yaml "github.com/goccy/go-yaml"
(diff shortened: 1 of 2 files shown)


mikefarah:
That is very interesting - I'm looking to replace the YAML library with goccy - and it seems like this a closer step towards that👍🏼  I'll have to have a closer when I get some time. The test isn't to standard - but let me first think on whether I want this merged 🤔

LarsStegman:
Hi @mikefarah do you already know whether you want to use Goccy instead of standard lib? Especially now that that one is archived.

mikefarah:
yaml/go-yaml now has support 🎆   ; closing this PR

LarsStegman:
@mikefarah there are still people who depend on goccy, so I would argue this PR is still relevant.

Issue #1441, opened 2022-11-21
Support of INI files


rkarpuzov:
Currently " -p p " looks similar syntax, but my tests failed with application.ini (firefox)"

Original file with:
````
#filter substitution
[App]
Vendor=MozillaTest
Name=LayoutDebug
Version=0.1
BuildID=@BUILD_ID@
Copyright=Copyright (c) 2004 Mozilla.org
ID={da915c15-c21a-41e2-95c3-2e0f76fd3191}
```

After change with  yq -i -p p '.Name = "Rossen"' application.ini
result is:

````

'[App]': ""
Vendor: MozillaTest
Name: Rossen
Version: "0.1"
BuildID: '@BUILD_ID@'
Copyright: Copyright (c) 2004 Mozilla.org
ID: '{da915c15-c21a-41e2-95c3-2e0f76fd3191}'

```'

After send run of the same command I receive more quotes in the result.

Best regards,
Rossen

mikefarah:
yeah nice idea - I'm working on toml support at present, after I get that working I may work on this. Happy to receive PRs of course :)

cz172638:
any update on it?
for me it looks similar to toml and properties

Zash:
This can be closed now?

rkarpuzov:
Yes!

In the test scenario the comment line is missing after processing, but this is subject of other discussion to keep the comments or no. 
INI support works for me and this is great job! 
Thank you all guys, who release this for me and the community!

PR #2375, opened 2025-05-16
Add INI support

b17k0:
https://github.com/mikefarah/yq/issues/1441


mikefarah:
Thank you! PR looks good - just need to run `make local test` and ensure everything passes :) You need to add 'noini' to project-words.txt so the spell check passes :)

b17k0:
@mikefarah Thanks! I've made the necessary updates, including adding 'noini' to project-words.txt. Now `make local test `runs without any errors, and everything should be in order. Let me know if there's anything else I should adjust!

Issue #2298, opened 2025-02-21
Replace go-yaml/yaml with goccy/go-yaml


neiser:
**Please describe your feature request.**
`yq` increasingly suffers from the fact that the underlying `go-yaml/yaml` library is rather unmaintained and lacks some important functionality for some `yq` use cases. 

**Describe the solution you'd like**
`yq` could migrate to a new kid in town, the [goccy/go-yaml](https://github.com/goccy/go-yaml). 
According to its README, the library aims to replace the `go-yaml/yaml` library and has some really [nice additional features](https://github.com/goccy/go-yaml?tab=readme-ov-file#features), such as:

> Reversible transformation without using the AST for YAML that includes Anchors, Aliases, and Comments

This feature is in particular interesting for `yq -i` use cases, which unfortunately might modify whitespace due to limitations in the `go-yaml/yaml` library.

mikefarah:
Totally agree, I actually started doing this late 2023 but hit some [snags](https://github.com/goccy/go-yaml/issues/404) on their repo. I can see That this was closed recently..worth trying again when I get free time

neiser:
@mikefarah Thanks for your effort! Really appreciate your effort!

goccy:
@mikefarah  Please let me know if the migration is being hindered by an issue with my library. I will prioritize addressing it.

Mivr:
Hey @mikefarah, will you have time for this soon?

And if not, can you offer some general guidance in case I or anyone else finds the time to work on this?

ccoVeille:
I would say it's urgent to wait for the fork that the YAML organization is now officially supporting

https://github.com/yaml/go-yaml

yesudeep:
I believe the existing code for go-yaml can stick around but we should migrate to an implementation fixes issues such as https://github.com/mikefarah/yq/issues/2294. Why not go ahead with the goccy parser and revisit if they bring the implementation up to speed? It would unblock the rest of us that are waiting for this to be fixed in the meantime.

stevenwdv:
Maybe this would fix #2386

mikefarah:
Def keeping my eyes on https://github.com/yaml/go-yaml - they haven't actually made any changes or releases afaik. I've just raised an issue against the repo, lets see how it goes.

Re goccy; I'm going to dedicate some time to that in June/July; I've been raising some issues here and there on goccy they are getting fixed which is great :)  ; just have to keep going through the tests and make sure it's not going to cause issues.

[...]

PR #2389, opened 2025-06-04
#2298 change yaml parser to a maintained one

Mivr:
Hello,

I did some work on changing the parser to goccy yaml instead of the unmaintained one, tests are passing and it seems to be working.

Do have in mind that goccy yaml does not preserve the order of fields, so this we can either fix in yq or consider a breaking change.

The current implementation is behind a flag so to not break users.

Note: I am a somewhat skilled go developer, but I haven't touched in years. Most of the code in this PR is generated with AI and then manually checked by me.

Fixes: #2298

### DIFF
--- a/cmd/constant.go
+++ b/cmd/constant.go
@@ -35,3 +35,5 @@ var completedSuccessfully = false
 var forceExpression = ""
 
 var expressionFile = ""
+
+var yamlParser = ""
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -94,10 +94,14 @@ yq -P -oy sample.json
 			logging.SetBackend(backend)
 			yqlib.InitExpressionParser()
 
-			// Handle legacy parser flag
-			useLegacyYamlParser, _ := cmd.Flags().GetBool("yaml-v3")
-			if useLegacyYamlParser {
+			// Handle YAML parser selection with validation
+			switch yamlParser {
+			case "goccy", "":
+				yqlib.ConfiguredYamlPreferences.UseGoccyParser = true
+			case "v3":
 				yqlib.ConfiguredYamlPreferences.UseGoccyParser = false
+			default:
+				return fmt.Errorf("invalid yaml-parser value '%s'. Valid options are: 'goccy', 'v3'", yamlParser)
 			}
 
 			return nil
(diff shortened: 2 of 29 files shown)


ccoVeille:
I would say it's urgent to wait for the fork that the YAML organization is now officially supporting

https://github.com/yaml/go-yaml

Mivr:
Oh, in that case, maybe a PR to remove goccy intial integration and use the new fork is relevant? (of course once the package is published)

And maybe I will make a PR to them for the case of "?" handling in go-yaml.

ccoVeille:
Here,I would let a maintainer reply

But I wanted to raise to anyone attention that yaml/go-yaml exists

mikefarah:
Good to see that go-yaml is getting support again - will be interesting to see how much activity it gets and how many issues they fix over the next couple of months. 

I'm not too sure on this PR - I'm currently working on the goccy decoder/encoder (albeit slowly because life 😅 ) would be good to get it done; but there's a lot of code here I'll need to work through.

Not sure about your comment of not preserving order - that would definitely be a deal breaker; but as far as I can tell it does preserve order.

Nonetheless, I'll be going on an extended holiday soon; will look at this (and progression of go-yaml) when I get back. Till then I won't be making fundamental changes or decisions on yq :)

mikefarah:
Btw - to _really_ know if the decoding/encoding work properly, you'll need to change the default encoder and decoder in operator.test (e.g. https://github.com/mikefarah/yq/blob/master/pkg/yqlib/operators_test.go#L56) to see if goccy can handle all the various scenarios

Mivr:
I see that go-yaml might have a version soon, so maybe keeping Goccy behind a flag is a good maybe short term solution as this will unblock some other issues like this one: https://github.com/aspect-build/rules_js/issues/2100

Mivr:
@mikefarah I think I fixed most of it, a lot of tests had to be ignored, do say what you prefer to do with the different parsing in goccy, for now its just an ignore but I can expect a better approach is needed

mikefarah:
Having tests ignored will produce more problems than it will solve

mikefarah:
Oh you've just updated the test expactation to match what the AI code generated :( Sorry but this is never going to work like that.

Mivr:
Hey @mikefarah I saw your changes in the main branch, and I see you are working on the goccy integration.

I also noticed you have a different approach to some of the things there. That is good and I am willing to work with your approach.

I opened a draft PR: https://github.com/mikefarah/yq/pull/2392

I would be happy on more feedback on this, at the moment there are about 18 cases where the parsing is not working as before. I will prepare a decent explanation on why and offer some options in the other pr.

The feedback from your last comment is addressed and no tests are changed, some had to be ignore, more on that in the other PR.

My goal is to make this work properly and I will be fixing any errors from the AI or mine. As I mentioned I am a decent developer, haven't written much golang in the last 7 years but I do have very solid foundations in about 11 languages.

My situation is that I am working on other projects that use yq, thus for me now it is a priority to fix yq parsing or replace it in the upstream projects. I hope we can work on this change with you and get it done sooner and with a quality you will be happy with.

PR #2392, opened 2025-06-07
#2298/change yaml parser to a maintained one

Mivr:


cmd/root.go
```diff
@@ -94,6 +94,16 @@ yq -P -oy sample.json
 			logging.SetBackend(backend)
 			yqlib.InitExpressionParser()
 
+			// Handle YAML parser selection with validation
+			switch yamlParser {
+			case "goccy", "":
+				yqlib.ConfiguredYamlPreferences.UseGoccyParser = true
+			case "v3":
```
ccoVeille: I would call it "legacy-v3", or simply "legacy"

It should be obvious this option is no longer maintained

ccoVeille (commented):
Minor remark, more from an end user than a real code review

cmd/root.go
Mivr: Good note, thanks will update

pkg/yqlib/datetime_preprocessor.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// Try to parse date-only format
+	if _, err := time.Parse("2006-01-02", value); err == nil {
+		return true
+	}
+
+	// Try RFC3339 without timezone
+	if _, err := time.Parse("2006-01-02T15:04:05", value); err == nil {
+		return true
+	}
```
ccoVeille: There are constants in time package for this.

You can use usestdlibvars linter to identify them easily

pkg/yqlib/datetime_preprocessor.go
```diff
@@ -0,0 +1,191 @@
...
+func isSkippableLine(line string) bool {
+	trimmed := strings.TrimSpace(line)
+
+	// Skip empty lines, comments, directives, document separators
+	if trimmed == "" || strings.HasPrefix(trimmed, "#") ||
+		strings.HasPrefix(trimmed, "%") || strings.HasPrefix(trimmed, "---") ||
+		strings.HasPrefix(trimmed, "...") {
+		return true
+	}
+
```
ccoVeille: I would suggest a switch with one case (or multiple) it would be more readable

pkg/yqlib/datetime_preprocessor.go
```diff
@@ -0,0 +1,191 @@
...
+	if strings.Contains(line, "!!") {
+		return true
+	}
+
+	// Skip multi-line scalar indicators
+	if strings.HasSuffix(trimmed, "|") || strings.HasSuffix(trimmed, ">") ||
+		strings.HasSuffix(trimmed, "|-") || strings.HasSuffix(trimmed, ">-") ||
+		strings.HasSuffix(trimmed, "|+") || strings.HasSuffix(trimmed, ">+") {
+		return true
+	}
```
ccoVeille: Same here

pkg/yqlib/datetime_preprocessor.go
```diff
@@ -0,0 +1,191 @@
...
+	trimmedValue := strings.TrimSpace(afterColon)
+
+	// Skip if value is empty, quoted, or already complex
+	if trimmedValue == "" ||
+		strings.HasPrefix(trimmedValue, "\"") ||
+		strings.HasPrefix(trimmedValue, "'") ||
+		strings.HasPrefix(trimmedValue, "{") ||
+		strings.HasPrefix(trimmedValue, "[") ||
+		strings.HasPrefix(trimmedValue, "&") ||
+		strings.HasPrefix(trimmedValue, "*") {
```
ccoVeille: Using a switch on trimmedValue[0] would be more readable

pkg/yqlib/datetime_preprocessor.go
```diff
@@ -0,0 +1,191 @@
...
+		return line
+	}
```
ccoVeille: Same here

pkg/yqlib/datetime_preprocessor.go
```diff
@@ -0,0 +1,191 @@
...
+func (dtp *DateTimePreprocessor) processKeyValueLine(line string, colonIndex int) string {
+	beforeColon := line[:colonIndex]
+	afterColon := line[colonIndex+1:]
+
+	// Check if the value part looks like a datetime
+	trimmedValue := strings.TrimSpace(afterColon)
+
+	// Skip if value is empty, quoted, or already complex
+	if trimmedValue == "" ||
+		strings.HasPrefix(trimmedValue, "\"") ||
```
ccoVeille: String literals notation can help

```suggestion
		strings.HasPrefix(trimmedValue, `"`) ||
```

pkg/yqlib/decoder_goccy_yaml.go
```diff
@@ -1,49 +1,238 @@
...
 }
 
-func NewGoccyYAMLDecoder() Decoder {
-	return &goccyYamlDecoder{}
+// processReadStream pre-processes the input stream to extract leading comments,
+// directives, and document separators before the main YAML content is parsed.
+// It returns a reader for the remaining stream and the extracted leading content.
+func (dec *goccyYamlDecoder) processReadStream(reader *bufio.Reader) (io.Reader, string, error) {
+	var commentLineRegEx = regexp.MustCompile(`^\s*#`)
+	var yamlDirectiveLineRegEx = regexp.MustCompile(`^\s*%YA`) // e.g., %YAML
```
ccoVeille: Move this out, otherwise the regexp is computed each time the function is called

pkg/yqlib/lib.go
```diff
@@ -116,14 +116,34 @@ func parseSnippet(value string) (*CandidateNode, error) {
...
+	if lhs == nil || rhs == nil {
+		return false
+	}
+
+	// Simple alias resolution - if one is alias and other isn't, resolve the alias
+	if lhs.Kind == AliasNode && rhs.Kind != AliasNode {
+		if lhs.Alias != nil && lhs.Alias.Alias != nil {
+			return recursiveNodeEqual(lhs.Alias.Alias, rhs)
+		}
+		return false
```
ccoVeille:
```suggestion
		if lhs.Alias == nil || lhs.Alias.Alias == nil {
			return false
		}
		return recursiveNodeEqual(lhs.Alias.Alias, rhs)
```

pkg/yqlib/datetime_preprocessor.go
```diff
@@ -0,0 +1,191 @@
...
+
+// ISO8601 date/time patterns that should be automatically tagged as timestamps
+var dateTimePatterns = []*regexp.Regexp{
+	// RFC3339 / ISO8601 with timezone: 2006-01-02T15:04:05Z or 2006-01-02T15:04:05+07:00
+	regexp.MustCompile(`^\s*([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(?:\.[0-9]+)?(?:Z|[+-][0-9]{2}:[0-9]{2}))\s*$`),
+	// Date only: 2006-01-02
+	regexp.MustCompile(`^\s*([0-9]{4}-[0-9]{2}-[0-9]{2})\s*$`),
+	// RFC3339 without timezone: 2006-01-02T15:04:05
+	regexp.MustCompile(`^\s*([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(?:\.[0-9]+)?)\s*$`),
+}
```
ccoVeille: I would move this aside the function that use it

Mivr:
@ccoVeille Thank you very much for the review. I will be addressing the comments tomorrow. 

And for the parser change, Goccy handles comments differently and this will break workflows for some users. I managed to equalize the behavior around timestamps and dates, but the comments will brake stuff. Also some arrays and anchors are handled differently, which will change the print out but the structure will be kept. So logical flows should not be impacted in contrast to the comments difference where if someone depends on some comment being attached to a specific node, his flow will break with the Goccy parser.

I will make a bit longer explanation of all the differences that I found and how they are addressed in this PR. Of course I am very opened for suggestions and comments. I will be working on this this and next week.

mikefarah:
You're racing ahead here, I'd slow this down a little - the go-yaml parser now has the support of the official Yaml org - I've been chatting with them and they will be soon addressing the issues raised. I'd prefer to keep go-yaml and maintain comments, anchors, array and formatting etc.

cmd/root.go
Mivr: Fixed

pkg/yqlib/datetime_preprocessor.go
Mivr: Surprisingly usestdlibvars did not show any errors, maybe my setup has a problem... 

Anyway fixed (next commit)

pkg/yqlib/datetime_preprocessor.go
Mivr: Fixed

pkg/yqlib/datetime_preprocessor.go
Mivr: Fixed

pkg/yqlib/datetime_preprocessor.go
Mivr: Fixed

pkg/yqlib/decoder_goccy_yaml.go
Mivr: Fixed

pkg/yqlib/datetime_preprocessor.go
Mivr: Fixed

Mivr:
@mikefarah 
## Thesis
You do make a good point: As I get it: Why to work on this if we can just wait for [https://github.com/yaml/go-yaml](https://github.com/yaml/go-yaml) and then just replace a dependency (theoretically).

I can state an argument for having a goccy parser as an alternative behind a flag:
- go-yaml has been unmaintained for years, so the actual backlog of issues that need addressing will be probably more than even the one in the now archived repo (~300 open issues). Assuming the YAML foundation continues working on this project with more or less the same speed as the last few months (~10 PRs merged for 2 months) they will have to work for years before they catch up on all the issues goccy has already solved.
    - In short my first point is that go-yaml even maintained by a very serious company is still years behind other implementations. So the time for getting a reliable parser is not the time it takes to release the first version of go-yaml, but the time it takes to fix a 3+ year legacy codebase with >300 issues opened.
- goccy can easily work behind a flag with a warning for known and possibly unknown incompatibilities, anyway the people that cannot use go-yaml and need goccy don't have much choice. They are either not working with yq because of go-yaml or they use goccy with some differences to standard yq.
- goccy discrepancies are not so many nor very intrusive (more details below).

## go-yaml vs goccy parsing differences (not mitigated already in this PR)

The differences between go-yaml and goccy displayed visually:
### 1. Comments association on array elements (BREAKING change if someone depends on comments inside the yaml file)
Can be mitigated with a pre processor that moves the comments so the user gets the same result in the end.

Input:
```YAML
name:
  # under-name-comment
  - first-array-child
```
   In this example go-yaml associates the comment with the first element of the array, while goccy associates the comment with the array itself

### 2. Flow maps will be expanded:
For input:
```YAML
{a: 1, b: 2}
```
go-yaml produces the same as output.


goccy produces:
```YAML
a: 1
b: 2
```
No workflows will be broken as the data structure is the same, formatting is the only change here.

It can be handled in some complicated pre + post processing but fairly I don't see this as a braking change, no one should depend on the formatting of the output yaml.

### 3. Merge anchor
Given:
```YAML
foo: &foo
  a: foo_a
  thing: foo_thing
  c: foo_c

foobar:
  c: foobar_c
  <<: *foo
  thing: foobar_thing
```

go-yaml output for .foobar:
```YAML
c: foobar_c
!!merge <<: *foo
thing: foobar_thing
```

Goccy output for .foobar:
```YAML
<<:
    a: foo_a
    c: foo_c
    thing: foo_thing
c: foobar_c
thing: foobar_thing
```

For me goccy seems to be the cleaner more informative output, again if you really want to this can be handled in post processor. 
### 4. Duplicate keys
go-yaml allows for duplicate keys
goccy does not

I think this is a feature rather than a bug. Again pre-processing can be used to level the behavior if this is what you believe is best.

### 5. Goccy is stricter on some invalid YAML cases
I think this is a feature.

## Conclusion
I do think goccy should be supported behind a flag for the use cases that have been waiting for years for a fix in go-yaml and the same will have to wait for years more even with the support of the YAML foundation.

Not implementing goccy as a parser or any other parser than go-yaml forces anyone that want to use yq to fix go-yaml and wait for their changes to be merged, which can take months or years even if they have the capacity to fix go-yaml.

More or less a parallel parser can save hundreds to thousands of work hours for the users of yq.

Mivr:
P.S. if you see value in having goccy as a parser and you want some or all of the differences to be handled with pre and/or post processors please do write so and I will add them to this PR.

LarsStegman:
> ### 4. Duplicate keys
> go-yaml allows for duplicate keys goccy does not

There is a `parser.Option` for goccy to allow duplicate keys, `parser.AllowDuplicateMapKey()`.

mikefarah:
I'm not convinced sorry - all this will do is double the maintenance that I need to do 😮‍💨  doubtless there will be niche issues for how goccy handles whitespace/comments etc as well as nich issues for go-yaml; and there will be confusion to users as to which parser to use when.

I'm only going to support a single yaml parser; and if go-yaml is going to get official support, and updated regularly then I'm prepared to stick with that.

Mivr:
So in the end the pnpm 10 issue was actually a bug in pnpm itself that generated a hard to parse yaml, after update of pnpm rules_js does not need yq to be able to parse the hard yaml.

And for yq I see your point. Seeing your last response time is a month I can guess that you don't have time to maintain this project, even with one library.

Anyway, my focus was rules_js to work well with pnpm v10 and as this is done, it is hard for me to justify working more on this. So if someone wants to use the changes of course it is free to do so.

Apart from this, as an alternative solution I made a [yaml2json](https://github.com/SimpleSoftwareOrg/yaml2json) tool for rules_js (as rules_js actually needs just that from yq) so maybe rules_js will migrate to it as it is quite faster than yq in initial testing.

With that I will close this draft PR.

Thanks to everyone that helped, your feedback was very valuable and I am very happy to have worked with you on this.

Issue #2110, opened 2024-07-23
!!merge overwrites existing keys


majewsky:
**Describe the bug**

Upon switching from python-yq to go-yq, I discovered that some existing YAML files are interpreted differently. Specifically, the `<<` merge key is handled differently by go-yq from what [the spec](https://yaml.org/type/merge.html) suggests. The spec says:

> If the value associated with the key is a single mapping node, each of its key/value pairs is inserted into the current mapping, unless the key already exists in it.

As shown in the example below (specifically, in the object called "ellipse"), go-yq appears to overwrite existing keys during merging. The object called "egg" demonstrates a workaround, in which the merge key is placed before all other keys to avoid this bug.

Version of yq: 4.44.2
Operating system: Arch Linux
Installed via: System package (pacman)

**Input Yaml**
```yaml
objects:
  - &circle
    name: circle
    shape: round
  - name: ellipse
    !!merge <<: *circle
  - !!merge <<: *circle
    name: egg
```

**Command**
The command you ran:
```
yq -o json input.yaml
```

**Actual behavior**

```yaml
{
  "objects": [
    {
      "name": "circle",
      "shape": "round"
    },
    {
      "name": "circle",
      "shape": "round"
    },
    {
      "shape": "round",
      "name": "egg"
    }
  ]
}
```

**Expected behavior**

```yaml
{
  "objects": [
    {
      "name": "circle",
      "shape": "round"
    },
    {
      "name": "ellipse",
      "shape": "round"
    },
    {
      "shape": "round",
      "name": "egg"
    }
  ]
}
```

mikefarah:
Ooh interesting 🤔  I don't know how I missed that. It's a pity they didn't have an example like yours, where the merge key is after the key values - I didn't think of it 😮‍💨 

I'm a little worried that if I change this behavior now it would break a bunch of data pipelines for current users - I don't think this bug can be safely fixed. 

I think that this is probably one of the reasons the merge key has been removed from the 1.2 spec :/ without reading the spec super clearly it's easy to have different expectations on how it would work :(

LarsStegman:
This is definitely a bug that needs to be fixed. I just ran into it and couldn't understand what was happening for hours. The behaviour is wrong, given the YAML standard.

mikefarah:
I've added a flag `--yaml-fix-merge-anchor-to-spec` which will be available in the next release. This will first default to false to maintain existing behavior (with a warning log).

[...]

PR #2400, opened 2025-06-13
Fix some issues around merge key/anchor handling

### LINKED ISSUE #2178: Explode with merge aliases throw an error
**Describe the bug**
I'm trying to work with [anchors and alias operators](https://mikefarah.gitbook.io/yq/operators/anchor-and-alias-operators) but I'm getting an error `merge anchor only supports maps, got !!seq instead` while I'm exploding the yaml.

It seems the line `- <<: *my-anchor` is the problem since I can work without that line but if it exists then throw the error.

Docs say  
> yq supports merge aliases (like <<: *blah) ...

So I didn't expected the error.

Version of yq: 4.44.3
Operating system: linux
Installed via: binary release

**Input Yaml**
```yaml
level_1:
- name: one
  level_2: &my-anchor
    - name: one_v1
      level_3:
        - name: foo
- name: two
  level_2:
    - <<: *my-anchor # uses my-anchor configuration
    - name: one_v1
      level_3: # add level_3.name: bar to my-anchor
        - name: bar

```

**Command**
```
yq 'explode(.)' data1.yml
```

**Actual behavior**

It throws an error:

```bash
Error: merge anchor only supports maps, got !!seq instead
```

**Expected behavior**

Just expand yaml and generate something like this:

```yaml
---
level_1:
  - name: one
    level_2:
      - name: one_v1
        level_3:
          - name: foo
  - name: two
    level_2:
      - name: one_v1 # this is the anchor property
        level_3:
          - name: foo
      - name: one_v1
        level_3:
          - name: bar
```

I've tried using other linters but format seems ok and could be possible to expand.  
Also I've tried to expand without `- <<: *my-anchor` line and not throw an error and I've run `yq '.level_1[0] | explode(.)' data1.yml` and also works fine. So problem seems to be the definition `- <<: *my-anchor`.

### LINKED ISSUE #2386: Merge key with inline map broken
**Describe the bug**
When I use a merge key (`<<:`) with an inline map rather than an anchor, `yq` doesn't propagate keys to the parent map.

Version of yq: 4.45.4
Operating system: linux
Installed via: binary release

**Input Yaml**
data1.yml:
```yaml
dummy: 42
<<:
  foo: 123
```

**Command**
```
yq eval '.bar' data1.yml
yq eval 'explode(.)' data1.yml
```

**Actual behavior**

`.bar`:
```yaml
null
```

`explode(.)`:
```yaml
dummy: 42
```

**Expected behavior**

`.bar`:
```yaml
123
```

`explode(.)`:
```yaml
dummy: 42
foo: 123
```

**Additional context**
Other YAML parsers seem fine with this, e.g. PyYAML (and online tools: https://www.yamllint.com/, https://jsonformatter.org/yaml-to-json). See also the [spec](https://yaml.org/type/merge.html).

I encountered this problem when implementing a custom `!include` tag using `load(.)`, where exploding `<<: !include "file.yml"` wouldn't work.

stevenwdv:
This PR fixes multiple issues related to merge keys (`!!merge <<`):

- Allow inline maps instead of just aliases (e.g. `{<<: {a: 42}}`, `{<<: [{a: 42}]}`), both for traversing and exploding (closes #2386)
- Allow aliases to sequences to be used (e.g. `{s: &s [{a: 42}], m: {<<: *s}}`), both for traversing and exploding (closes #2178)
- Disallow nested sequences (e.g. ~~`{<<: [[{a: 42}]]}`~~)
- Disallow other (inline) types (e.g. ~~`{<<: 42}`~~)
- However: Ignore invalid merge keys when traversing maps. Previously cases like `{a: &a 42, m: {<<: *a}}` were reported but cases like `{m: {<<: 42}}` were not. I chose to ignore all errors because otherwise traversing such a map (e.g. to correct types) becomes impossible.
- Errors are *not* ignored when using `explode`. Previously explode would just leave out invalid merge keys. Another potential solution would be to leave them as-is.
- Fix precedence of merge anchor sequence for traverse (explode was already correct), now it correctly lets earlier keys overwrite later keys according to the spec.
- Fix excessive exploding with merge keys, e.g. previously `echo '{b: &b {a: &a 42}, r: *a, c: {<<: *b}}' | yq 'explode(.c)'` would yield `{b: &b {a: 42}, r: *a, c: {a: 42}}`, making `r: *a` a dangling reference. Now it yields `{b: &b {a: &a 42}, r: *a, c: {a: &a 42}}`.
- For exploding, detect merge keys using tag (`!!merge`) rather than value (`<<`).

I do still have a question: I saw in `operator_traverse_path_test.go` that goccy would reject at least some invalid merge types at parse time, which might make using custom tags like `<<: !include file.yaml` (see #2386) impossible. Am I right about that? And does that mean that this usage will not be supported in the future? If so, that would be a bit sad. See also the note above about ignoring invalid merge keys.


pkg/yqlib/operator_traverse_path_test.go
```diff
@@ -553,9 +569,16 @@ var traversePathOperatorScenarios = []expressionScenario{
...
-		expectedError: "can only use merge anchors with maps (!!map), but got !!seq",
+		expectedError: "can only use merge anchors with maps (!!map) or sequences (!!seq) of maps, but got sequence containing !!str",
 		skipForGoccy:  true, // throws an error on parsing, that's fine
 	},
+	{
+		skipDoc:       true,
+		document:      `{<<: 42}`,
+		expression:    ".[]",
+		expectedError: "can only use merge anchors with maps (!!map) or sequences (!!seq) of maps, but got !!int",
+		skipForGoccy:  true,
```
stevenwdv, line 580: I didn't try to test goccy but I assume it will reject this if it rejects the case above

stevenwdv:
Wait, this actually doesn't fix `explode(.)` yet. Converting to draft for now, I'll do that later.

stevenwdv:
**EDIT:** nvm fixed it

---
Ok, the excessive exploding fix I mentioned above may actually cause problems of its own:
`echo '{a: {b: &b 42}, <<: {c: *b}}' | yq 'explode(.)'` now returns `{a: {b: 42}, c: *b}`, which also has a dangling alias...

However, I don't really know how to solve this, because it depends on if the mapping node to which the merge anchor refers is also being exploded, e.g. with `{r: &r {...}, m: {<<: *r}}` we don't want to explode `.r`, although we might make a copy and then explode it.

And apparently this issue is already present when *partially* exploding: `echo '{a: {b: &b 42}, c: *b}' | yq 'explode(.a)'` in the latest release v4.45.4 returns `{a: {b: 42}, c: *b}`, which also has a dangling alias. Is that a bug or intentional?

pkg/yqlib/operator_anchors_aliases_test.go
```diff
@@ -283,6 +283,66 @@ var anchorOperatorScenarios = []expressionScenario{
 		expression:     `.thingOne |= explode(.) * {"value": false}`,
 		expected:       []string{expectedUpdatedArrayRef},
 	},
+	{ // Merge anchor with inline map
```
mikefarah: put comments as the description instead, they will then be printed on test failure for traceability

mikefarah:
Minor comment, need more time to review this and consider if there are any side effects - thanks for the contribution though!

pkg/yqlib/operator_traverse_path_test.go
```diff
@@ -425,11 +449,11 @@ var traversePathOperatorScenarios = []expressionScenario{
 	},
 	{
 		description:    "Traversing merge anchor lists",
-		subdescription: "Note that the later merge anchors override previous",
+		subdescription: "Note that the keys earlier in the merge anchors sequence override later ones",
 		document:       mergeDocSample,
 		expression:     `.foobarList.thing`,
 		expected: []string{
-			"D0, P[bar thing], (!!str)::bar_thing\n",
+			"D0, P[foo thing], (!!str)::foo_thing\n",
```
mikefarah: Oooh nice find - this isn't to the YAML spec and it's works opposite to explode 🤦🏼  _but_ this is a breaking change... and will cause all sorts of issues for people :/ Hmmmm

mikefarah: I'm happy to go through with this PR if either; this fix is not included, or it's controlled via a new flag, maybe `--yaml-correct-merge-order` or something, defaulting to false. Should have a warning in there if the flag is not set and the logic is executed.

Then later with enough notice given, I can update the default flag value to true.

stevenwdv: A flag may be a good idea, I'll have a look at it later. Maybe the same could be done for #2110.

mikefarah: Yeah good idea 👍🏼

mikefarah: I think it could be done under the same flag, I'm happy to work on 2110 and adding a flag, then perhaps you can add the `if` check into this PR?

mikefarah: I've added a flag `--yaml-fix-merge-anchor-to-spec` available as a check in code as `ConfiguredYamlPreferences.FixMergeAnchorToSpec`. If you can add this for the fix you've done in traverse that'd be ace 👍🏼  You can see how I've tested true/false in the operator_anchor_aliases_test.go

stevenwdv: Nice! I put the fixed overriding behind the flag. You missed fixing #2110 for traversing (only exploding), but I could fix it both for local keys (#2110) & list merge keys at once anyway.

I also noticed that after 6e8cc00030b8165f159d72cc39ccfd67a97c483d it'll warn about overriding behavior with list merge anchor keys as well (e.g. `{<<: [{a: 1}, {a: 2}]}`), while this behavior is and was already correct (`explode(.).a` = `1`), see the `TODO`s I added. Maybe you know of a good fix? Or add 'unless...' to the warning..?

stevenwdv: Oh, I just remembered, in the main branch the flag currently actually *breaks* list merge anchors for explode... 🙃

Edit: added a testcase for this

stevenwdv: I just realized that the way I fixed #2110 for `operator_traverse_path` actually breaks overriding keys in regular maps... :( I added a failing test.

I really wonder what the rationale behind the ordering in the merge keys spec was..

pkg/yqlib/operator_anchors_aliases.go
```diff
@@ -237,42 +226,91 @@ func explodeNode(node *CandidateNode, context Context) error {
...
 
 	for newEl := newContent.MatchingNodes.Front(); newEl != nil; newEl = newEl.Next() {
 		valueEl := newEl.Next() // move forward twice
 		keyNode := newEl.Value.(*CandidateNode)
 		log.Debugf("checking new content %v:%v", keyNode.Value, valueEl.Value.(*CandidateNode).Value)
 		if keyNode.Value == key.Value && keyNode.Alias == nil && key.Alias == nil {
-			log.Debugf("overridign new content")
+			log.Debugf("overriding new content")
 			if !ConfiguredYamlPreferences.FixMergeAnchorToSpec {
+				//TODO This also fires in when an earlier element in a list merge anchor overwrites a later element, which *is* to the spec
```
mikefarah: ooh darn so this means I broke that behavior? I thought I would have had a failing test somewhere :/

pkg/yqlib/operator_anchors_aliases_test.go
```diff
@@ -122,12 +142,22 @@ var anchorOperatorScenarios = []expressionScenario{
 		expression:     ".[4] | explode(.)",
 		expected:       []string{"D0, P[4], (!!map)::r: 10\nx: 1\ny: 2\n"},
 	},
+	//TODO The following 2 tests warn about overwriting [3].r not being to spec while they shouldn't
```
mikefarah: Ah I see...thanks for making this super clear! Will fix...somehow

mikefarah:
This is really good - I think you have a clearer picture of how merge works better than I do 😅  - I'm going to go over it again later just to double check it as it's a sensitive part of the code.

stevenwdv:
Thanks for having another look! Did you see my comments in the [thread above](https://github.com/mikefarah/yq/pull/2400#discussion_r2210545616) as well? Currently I've actually broken key overriding in regular maps for traversing, and I don't have a good solution yet...

stevenwdv:
Ah spelling is also wrong I see, didn't know this was a bri'ish codebase ;)

Edit: apparently the codebase it not entirely consistent, but anyway

mikefarah:
Ok I think to properly fix explode I need to take a step further and rewrite the reconstructAliasedMap - which I've done in a branch  `MakeExplodeGreatAgain` and included your extra tests 👍🏼   I've also changed the warning to just be a generic you haven't set the flag warning, rather than trying to specify which maps are affected.

Tempted to merge it into master - it'll no doubt conflict with what you've got here though :/

stevenwdv:
> Ok I think to properly fix explode I need to take a step further and rewrite the reconstructAliasedMap - which I've done in a branch `MakeExplodeGreatAgain`

Nice, I'll merge it in here as well

stevenwdv:
@mikefarah I notice that `fixedReconstructAliasedMap` in `MakeExplodeGreatAgain` is lacking a lot of logic that is present in `reconstructAliasedMap`. Compare:
```
$ echo '{e: &empty {}, b: &b 42, a: {b2: *b}, <<: *empty}' | ./yq 'explode(.)'
[WARNING] 14:18:06 --yaml-fix-merge-anchor-to-spec is false; causing merge anchors to override the existing values which isn't to the yaml spec. This flag will default to true in late 2025.
{e: {}, b: 42, a: {b2: 42}}
```
vs
```
$ echo '{e: &empty {}, b: &b 42, a: {b2: *b}, <<: *empty}' | ./yq --yaml-fix-merge-anchor-to-spec 'explode(.)'
{e: &empty {}, b: &b 42, a: {b2: *b}}
```
This is because it's not using `overrideEntry`. It's also not using `Context`, but I'm actually not sure what the exact purpose of that is.

This may also break aliases (`*b` is invalid now):
```
$ echo '{b: &b 42, m: {e: &empty {}, a: {b2: *b}, <<: *empty}}' | ./yq --yaml-fix-merge-anchor-to-spec 'explode(.)'
{b: 42, m: {e: &empty {}, a: {b2: *b}}}
```


I'm seeing if I can quickly fix it in the merge.

stevenwdv:
> I'm seeing if I can quickly fix it in the merge.

I thought of a way to quickly solve the issue in a different way, by evaluating merge keys first, I hope you don't mind.

I'm now looking at the issue in my code I mentioned above.

stevenwdv:
Ok, it should be all done now, I hope! Well, maybe the tests could be cleaned up a bit / made more consistent between traverse & explode.

Btw, I thought of some interesting edge cases, where I don't know what the intended way to handle them would be, which again is probably one of the reasons merge anchors are not part of the recent spec anymore: the more I know about them, the less sense it makes to me to use regular keys for behavior like this.

(current behavior in **bold**)

- `{<<: {a: 42}, <<: {b: 42}}`: **`{a: 42, b: 42}`**/`{b: 42}`?
  Arguably it should be the latter, since after de-duplication only the second merge key remains
- `{<<: {a: 42}, !!str <<: {b: 42}}`: **`{a: 42, !!str <<: {b: 42}}`**/`{!!str <<: {b: 42}}`?
  I think this is right, since the keys have different types
- `{<<: {a: 42}, !!merge mykey: {a: 37, b: 42}}`: **`{a: 42, b: 42}`**/`{a: 37, b: 42}`/invalid?

Not that other YAML tools handle them in a good way; [yamllint.com](https://www.yamllint.com/) with input `{<<: {a: 42}, !!str <<: {b: 42}}` gives `{a: 42, b: 42}` (as did `yq` previously; with explode, not traverse).

mikefarah:
Ah the reason I was processing like that was to preserve key order on merging. Not a big issue though, this is looking good! Thanks for PR!

mikefarah:
Hmm - only problem with the way you're processing explode is that they key order is mangled if merge is processed first :/ Not a massive issues I guess...though I might see if I can fix that

stevenwdv:
True, although YAML does act like merge keys come first, so it kinda makes sense in a way

stevenwdv:
Hey @mikefarah, I'm really confused because it seems like something went really wrong merging this; a lot of changes didn't make it and some logic from `MakeExplodeGreatAgain` that I changed it in this MR was merged. E.g.
- `fixedReconstructAliasedMap` re-appeared
- `!!str <<` is broken again
- Inline merge anchors are broken without `--yaml-fix-merge-anchor-to-spec` while I fixed those for both cases
- Nested merge anchors (`{<<: {<<: {a: 42}}}`) are broken even with the fix
- Probably more

mikefarah:
Yeah, I brought those changes back in to ensure the correct map key order. Afaik I included all the tests you added; so there's shouldn't be any gaps :/ - but maybe something went awry for those cases. 

Inline merging is under the flag, that's intentional - my thinking is that it's best to leave the old way of merging as untouched as possible seeing the new way is a complete re-write.

Not sure what `!!str <<` exactly refers to? Will take a look at nested inline merge anchors when I get a chance - though it seems pretty niche; not sure what the use-case would be?

stevenwdv:
@mikefarah
I'm a bit sad to see many things that worked with this MR not work anymore. I put in a lot of work to try and get everything right, only to see most of it break again. Sorry if that sounds to harsh, I'm still really glad you're maintaining this useful tool and actively looking at issues and comments. Anyway, more specifically, it still doesn't work anymore for what I was trying to do originally, also with the flag. (Specifically because of nested inline merge anchors.)

I think that aside from the order, which I understand you might want to keep the same but shouldn't impact the meaning, I made sure to keep everything compatible with previous behavior with and without the flag. I also think that I kept the order the same as before when the flag was not specified (although I'm not 100% sure of the top of my head); maybe you can check this, and in that case, can't the other fixes just be included with and without the flag? I don't think those could break previous code.

Regarding behavior with the flag, `fixedReconstructAliasedMap` bypasses all original fixes from this MR, so it doesn't handle inline maps in merge anchors, and has some other shortcomings.

I think that not all tests are ran for both with & without the flag anymore where they could be. Maybe that's why it seems to be working fine.

`!!str <<` is not a merge key, it's just a regular key that happens to be called `<<`.

I could make a PR to fix the things that are broken now and add tests for those, but the thing is; this was that PR :/

PR #2421, opened 2025-07-11
skip format check for filenames ending with dot

ryenus:


mikefarah:
Nice, can you add a test as well please :)

ryenus:
I just gave it a try and luckily it worked: 
- https://github.com/mikefarah/yq/actions/runs/16237369374/job/45849259334?pr=2421#step:5:1

.github/workflows/go.yml
```diff
@@ -27,6 +27,11 @@ jobs:
             dep ensure
         fi
 
+    - name: Run Go tests in test folder
```
mikefarah: this already runs as part of make local build

ryenus: Done, I just removed it and squashed the commits.

ryenus:
@mikefarah build failing due to gosec download error:

```log
+ curl -w %{http_code} -sL -H Accept:application/json -o /tmp/tmp.MNRaKEukA0 https://api.github.com/repos/securego/gosec/releases/latest
+ code=403
+ [ 403 != 200 ]
+ log_debug http_download_curl received HTTP status 403
```

See https://github.com/mikefarah/yq/actions/runs/16279900703/job/45967251140?pr=2421#step:5:56

scripts/secure.sh
```diff
@@ -3,9 +3,15 @@
...
+
+command -v gosec &> /dev/null && GOSEC=gosec || GOSEC=./bin/gosec
+if [ -x "$GOSEC" ]; then
+    echo "Running Gosec security checks..."
+    "$GOSEC" "${PWD}" ./...
 else
-  ./bin/gosec "${PWD}" ./...
-fi
\ No newline at end of file
+    echo "Gosec not found, skipping security checks."
```
mikefarah: I don't like this change to how gosec runs, I prefer it to be the way it was before; where it runs (as close as possible) the same locally as it does in CI. Pretty sure the 403 problem was a temporary thing (master still passes...). This change adds more complexity, as well as it will inadvertently skip gosec if not present, which is not what I want. 

I think it'd be better to keep the PR change as small as possible - and not introduce any I/CD changes. If there is a gap in the CI/CD feel free to raise a separate PR and discuss there 👍🏼

ryenus: > I think it'd be better to keep the PR change as small as possible - and not introduce any I/CD changes. If there is a gap in the CI/CD feel free to raise a separate PR and discuss there

Totally agree, I just need a green build to be sure the original fix is OK. And that's why the CI part has been done with a separate commit, so that it can be dropped out.

ryenus:
Hey @mikefarah, I've dropped the CI change, build still failing though, nevertheless we know the code had been scanned and it was fine.

mikefarah:
Yeah no worries, I'll look at it on master, thanks @ryenus 👍🏼

PR #2450, opened 2025-08-21
Use ghcr.io instead of ghrc.io

MorrisLaw:
Fix typo that points to the wrong registry.

Fixes: #2451

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -95,6 +95,6 @@ jobs:
             -t "${IMAGE_NAME}:4-githubaction" \
             -t "${IMAGE_NAME}:latest-githubaction" \
             -t "ghcr.io/${IMAGE_NAME}:${IMAGE_VERSION}-githubaction" \
-            -t "ghrc.io/${IMAGE_NAME}:4-githubaction" \
+            -t "ghcr.io/${IMAGE_NAME}:4-githubaction" \
             -t "ghcr.io/${IMAGE_NAME}:latest-githubaction" \
             .


romanm-anyword:
Yeah, see https://bmitch.net/blog/2025-08-22-ghrc-appears-malicious/

MorrisLaw:
> Yeah, see https://bmitch.net/blog/2025-08-22-ghrc-appears-malicious/

Precisely! Thanks for adding the link for context.

MorrisLaw:
cc/ @mikefarah

sstelfox:
It doesn't look like this typo could have resulted in a security breach of this repository. The GITHUB_TOKEN is scoped to the correct registry, and is not granted any additional permissions in this workflow.

In the event the token did find its way to the malicious server, this repository was created March 8th, 2023 after the legacy GITHUB_TOKEN permissions expired (February 2, 2023 was the cut off). Without additional permissions the permissions available would have been `content:read` and `metadata:read`, neither of which could have allowed for changes or impact to the content of this repository.

mikefarah:
That's quite concerning - thanks for looking into this and raising a PR folks.

mikefarah:
I don't think this exposed anything either - the docker login (with the GITHUB_TOKEN) is for the correct domain ghrc.

Trying to find more info on that token - I don't have any fine grain or personal access tokens setup; so it'll be what Github automatically generates. If it was being passed to the bad domain (which I'm certain it wasn't because the script doesn't login against it); I wonder if it would have allowed someone with that credential to clobber the binary releases with something malicious

mikefarah:
Yeah from that linked page above:

```
Some scenarios that would result in credentials being leaked include:

    Running docker login ghrc.io.
    A GitHub action with uses: docker/login-action and with registry: ghrc.io.
    Creating a Kubernetes secret with registry credentials for ghrc.io and then trying to pull an image from that typoed host.

Simply trying to push or pull an image to this registry without logging in will not leak credentials and will not leak any data other than your repository name. These commands default to trying to acquire an anonymous token, which will quickly fail.
```

Does seem like a close call though 😥

thiagowfx:
[zizmor](https://docs.zizmor.sh/) is helpful to detect/prevent abuse vectors via github actions.

I did a sample run in this repo:

> 26 findings (9 suppressed, 7 fixable): 0 unknown, 1 informational, 0 low, 8 medium, 8 high

Example finding:

```
info[template-injection]: code injection via template expansion
  --> ./.github/workflows/docker-release.yml:31:23
   |
31 |         run: echo ${{ steps.buildx.outputs.platforms }} && docker version
   |         ---           ------------------------------ info: may expand into attacker-controllable code
   |         |
   |         info: this run block
   |
   = note: audit confidence → Low
   = note: this finding has an auto-fix
```

It does not detect this typo though. Could be a good FR to file upstream :)

mikefarah:
Nice - I'll give that a run through hopefully this weekend and see what it finds

PR #2458, opened 2025-09-01
feat: configure colors of output with env variables

hamskerpir:
Hey,
my usecase was that color that I was using for my terminal prompt is same that is hardcoded in `yq`, so I've decided add a bit of flexibility to the tool.
Unfortunately, I did not find a config structure, and if it does not exist, I think I'm not the person that should set such change in repository :D 

Anyway, here is my small changes to configure output via env vars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,6 +423,30 @@ Flags:
 
 Use "yq [command] --help" for more information about a command.
 ```
+
+## Environment Variables
+
+### Color Customization
+
+You can customize the colors used in YAML output by setting these environment variables:
+
+- `YQ_COLOR_BOOL` - Color for boolean values (default: hi-magenta)
+- `YQ_COLOR_NUMBER` - Color for numeric values (default: hi-magenta)  
+- `YQ_COLOR_MAP_KEY` - Color for map keys (default: cyan)
+- `YQ_COLOR_ANCHOR` - Color for YAML anchors (default: hi-yellow)
+- `YQ_COLOR_ALIAS` - Color for YAML aliases (default: hi-yellow)
+- `YQ_COLOR_STRING` - Color for string values (default: green)
+- `YQ_COLOR_COMMENT` - Color for comments (default: hi-black)
+
+Supported color values: `black`, `red`, `green`, `yellow`, `blue`, `magenta`, `cyan`, `white`, and their `hi-` prefixed variants (e.g., `hi-red`).
+
+Example:
+```bash
+export YQ_COLOR_STRING=red
+export YQ_COLOR_MAP_KEY=hi-blue
+yq -C '.data' file.yaml
+```
+
 ## Known Issues / Missing Features
 - `yq` attempts to preserve comment positions and whitespace as much as possible, but it does not handle all scenarios (see https://github.com/go-yaml/yaml/tree/v3 for details)
 - Powershell has its own...[opinions on quoting yq](https://mikefarah.gitbook.io/yq/usage/tips-and-tricks#quotes-in-windows-powershell)
--- a/pkg/yqlib/color_config.go
+++ b/pkg/yqlib/color_config.go
@@ -1,0 +1,93 @@
+package yqlib
+
+import (
+	"fmt"
+	"github.com/fatih/color"
+	"os"
+)
+
+type ColorConfig struct {
+	Bool    color.Attribute
+	Number  color.Attribute
+	MapKey  color.Attribute
+	Anchor  color.Attribute
+	Alias   color.Attribute
+	String  color.Attribute
+	Comment color.Attribute
+}
+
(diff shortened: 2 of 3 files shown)


hamskerpir:
image

pkg/yqlib/color_config.go
```diff
@@ -0,0 +1,117 @@
...
+
+type ColorConfig struct {
+	BoolColor    color.Attribute
+	NumberColor  color.Attribute
+	MapKeyColor  color.Attribute
+	AnchorColor  color.Attribute
+	AliasColor   color.Attribute
+	StringColor  color.Attribute
+	CommentColor color.Attribute
+}
```
ccoVeille: Nitpicking here.

I'm unsure there is a need for having a Color suffix in a ColorConfig struct

Hence I'm suggesting this

```suggestion
type ColorConfig struct {
	Bool    color.Attribute
	Number  color.Attribute
	MapKey  color.Attribute
	Anchor  color.Attribute
	Alias   color.Attribute
	String  color.Attribute
	Comment color.Attribute
}
```

README.md
```diff
@@ -415,6 +415,30 @@ Flags:
...
+
+- `YQ_COLOR_BOOL` - Color for boolean values (default: hi-magenta)
+- `YQ_COLOR_NUMBER` - Color for numeric values (default: hi-magenta)  
+- `YQ_COLOR_MAP_KEY` - Color for map keys (default: cyan)
+- `YQ_COLOR_ANCHOR` - Color for YAML anchors (default: hi-yellow)
+- `YQ_COLOR_ALIAS` - Color for YAML aliases (default: hi-yellow)
+- `YQ_COLOR_STRING` - Color for string values (default: green)
+- `YQ_COLOR_COMMENT` - Color for comments (default: hi-black)
+
+Supported color values: `black`, `red`, `green`, `yellow`, `blue`, `magenta`, `cyan`, `white`, and their `hi-` prefixed variants (e.g., `hi-red`). Numeric color codes are also supported.
```
ccoVeille: Please provide an example of the color codes format.

It's not clear if it's an RGB thing or the terminal color code like 39

pkg/yqlib/color_config.go
```diff
@@ -0,0 +1,117 @@
...
+		if attr, err := parseColorAttribute(colorStr); err == nil {
+			config.StringColor = attr
+		}
+	}
+
+	if colorStr := os.Getenv("YQ_COLOR_COMMENT"); colorStr != "" {
+		if attr, err := parseColorAttribute(colorStr); err == nil {
+			config.CommentColor = attr
+		}
+	}
```
ccoVeille: You could define a map or slice and iterate.

It's not logic to replicate same code so many times

pkg/yqlib/color_config.go
```diff
@@ -0,0 +1,117 @@
...
+		return color.FgHiBlue, nil
+	case "hi-magenta":
+		return color.FgHiMagenta, nil
+	case "hi-cyan":
+		return color.FgHiCyan, nil
+	case "hi-white":
+		return color.FgHiWhite, nil
+	default:
+		if num, err := strconv.Atoi(colorStr); err == nil {
+			return color.Attribute(num), nil
```
mikefarah: What does this do?

mikefarah:
I agree with @ccoVeille  comments 👍🏼

hamskerpir:
sorry for long inactivity

pkg/yqlib/color_config.go
hamskerpir: removed this

README.md
hamskerpir: removed this

pkg/yqlib/color_config.go
```diff
@@ -0,0 +1,93 @@
...
+		"YQ_COLOR_COMMENT": &config.Comment,
+	}
+
+	for envVar, configField := range colorMappings {
+		if colorStr := os.Getenv(envVar); colorStr != "" {
+			if attr, err := parseColorAttribute(colorStr); err == nil {
+				*configField = attr
+			}
+		}
+	}
```
ccoVeille, line 46: I tend to prefer code with less branches

```suggestion
	for envVar, configField := range colorMappings {
		colorStr := os.Getenv(envVar)
		if colorStr == "" {
			continue
		}

		attr, err := parseColorAttribute(colorStr)
		if err != nil {
			continue
		}
		*configField = attr
	}
```

Also, I feel like it could be frustrating for someone if they try to use an unsupported value, they won't know it failed. I would report the error to user

mikefarah:
I'm not convinced that this makes sense to do for yq - I think it'll just be additional complexity to maintain. Not sure how common it is to be able to control colours of CLI tools - but I haven't really seen a demand for it

PR #2487, opened 2025-10-01
fix: keep xml namespace prefixes for tags 

### LINKED ISSUE #1730: Xml: is keep the namespace
### Discussed in https://github.com/mikefarah/yq/discussions/1721

<div type='discussions-op-text'>

<sup>Originally posted by **elandorr** July  8, 2023</sup>
Hi

is it possible to keep the ns here?

```xml
<?xml version="1.0" encoding="UTF-8"?>
<foo>
    <title>barfoo</title>
    <macroni:title>foobar</macroni:title>
</foo>
```

The `macroni` is lost.

```yml
+p_xml: version="1.0" encoding="UTF-8"
foo:
  title:
    - barfoo
    - foobar
```

I have a file that uses such a format, which means this borks:

```
yq -pxml -oy '.foo.title' fooo.xml 

- barfoo
- foobar
```

Ideally it would only show the `<title>`.

Appreciate yq, really useful, thanks!</div>

baodrate:
fixes `--xml-keep-namespace=true` so that element tags keep their namespace prefix. e.g. 

```xml
<xsi:foo>bar</xsi:foo>
```

does a roundtrip unchanged (rather than becoming `<foo>bar</foo>`)

fixes #1730

### DIFF
--- a/pkg/yqlib/decoder_xml.go
+++ b/pkg/yqlib/decoder_xml.go
@@ -270,10 +270,16 @@ func (dec *xmlDecoder) decodeXML(root *xmlNode) error {
 			log.Debug("start element %v", se.Name.Local)
 			elem.state = "started"
 			// Build new a new current element and link it to its parent
+			var label = se.Name.Local
+			if dec.prefs.KeepNamespace {
+				if se.Name.Space != "" {
+					label = se.Name.Space + ":" + se.Name.Local
+				}
+			}
 			elem = &element{
 				parent: elem,
 				n:      &xmlNode{},
-				label:  se.Name.Local,
+				label:  label,
 			}
 
 			// Extract attributes as children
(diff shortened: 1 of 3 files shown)


baodrate:
Note that when using `--xml-raw-token=false` this causes any elements with a default namespace (i.e. the `xmlns` attribute) to become prefixed with the value of `xmlns`.

This might be out of scope of this PR since `--xml-raw-token=false` seems to behave incorrectly anyways. e.g.

`yq -px -ox --xml-raw-token=false` (prior to this PR) already processed:

```xml
<foo xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <bar xsi:abc="xyz">123</bar>
</foo>
```

into:

```xml
<foo xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <bar http://www.w3.org/2001/XMLSchema-instance:abc="xyz">123</bar>
</foo>
```

which is invalid xml.

I'm not sure how `xml-raw-token` is *supposed* to work, so I'm unsure how to fix this. But in any case, it should likely be tracked in a separate issue

Issue #2518, opened 2025-11-24
Exec format error in Arm64 systems in v4.49.1


ManojGuha:
**Describe the bug**
After the Nov 22, 2025 update, yq is not working as expected in Arm64 ( Nvidia Jetson Xavier NX ) systems. 
When I run "yq" in terminal , getting this error: "cannot snap-exec: cannot exec "/snap/yq/2747/bin/yq": exec format error".

Version of yq: v4.49.1
Operating system: Ubuntu 20.04 Arm64 ( Nvidia Jetson Xavier NX 5.10.104-tegra )
Installed via: Snap

**Command**
The command you ran:
```
yq
```

**Actual behavior**

```
cannot snap-exec: cannot exec "/snap/yq/2747/bin/yq": exec format error
```

**Expected behavior**

```
The output of yq command as follows: 
Usage:
  yq [flags]
  yq [command]

Examples:

# yq tries to auto-detect the file format based off the extension, and defaults to YAML if it's unknown (or piping through STDIN)
# Use the '-p/--input-format' flag to specify a format type.
cat file.xml | yq -p xml

# read the "stuff" node from "myfile.yml"
yq '.stuff' < myfile.yml

# update myfile.yml in place
yq -i '.stuff = "foo"' myfile.yml

# print contents of sample.json as idiomatic YAML
yq -P -oy sample.json

...

```

Additional info 

Output of command : file /snap/yq/2747/bin/yq

```
/snap/yq/2747/bin/yq: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, BuildID[sha1]=8ba0eb238439cb14695a73a5daa5fbceb6182059, with debug_info, not stripped
```

aalexjo:
I am also experiencing this issue. Looks like it comes from 8e86bdb876f84faaef3af6e9fd24db21b75192e8, where each architecture was build and published, but snapcraft does not crosscompile, causing incorrectly labeled binaries to be pushed. Launchpad provides native builders ({'riscv64', 'arm64', 's390x', 'ppc64el', 'amd64', 'armhf'}) when building with snapcraft [remote build](https://documentation.ubuntu.com/snapcraft/stable/explanation/remote-build/). I have made a PR that does this (unfortunately, this involves changes in the release pipeline @mikefarah) . 
I am not sure if there is a more direct way of addressing this using xcompile for snap and pushing that somehow.

ddickey-blitzoo:
[...]

PR #2519, opened 2025-11-24
Fixed architecture builders

aalexjo:
**What this PR does**

Reverts snap builds to use native build-on per architecture so each snap is compiled on the correct platform.
Fixes misbuilt arm snaps that were produced on amd64 and failed with exec format errors on arm devices.

**Why**
Snapcraft’s Go plugin doesn’t cross-compile; building arm snaps on amd64 produced amd64 binaries mislabeled as arm, breaking installs on arm. Using native builders per arch restores correct binaries for arm64/armhf (and others).

### DIFF
--- a/.github/workflows/snap-release.yml
+++ b/.github/workflows/snap-release.yml
@@ -17,17 +17,22 @@ on:
           - "false"
 
 jobs:
-    buildSnap:
-      environment: snap
-      runs-on: ubuntu-latest
-      steps:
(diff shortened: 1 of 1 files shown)


aalexjo:
This addresses #2518.

aalexjo:
@mikefarah We see devices that update snaps automatically are breaking, giving this a fairly high urgency.

PR #2520, opened 2025-11-24
Update architechtures

aalexjo:
**What this PR does**
Remote build snap builds to use native build-on per architecture so each snap is compiled on the correct platform.
Fixes misbuilt arm snaps that were produced on amd64 and failed with exec format errors on arm devices.
Build using launchpad remote builder instead of the github action worker.

**Why**
Snapcraft’s Go plugin doesn’t cross-compile; building arm snaps on amd64 produced amd64 binaries mislabeled as arm, breaking installs on arm. Using native builders per arch restores correct binaries for arm64/armhf (and others).

### DIFF
--- a/.github/workflows/snap-release.yml
+++ b/.github/workflows/snap-release.yml
@@ -15,6 +15,10 @@ jobs:
         - uses: actions/checkout@v5
         - uses: snapcore/action-build@v1
           id: build
+          env:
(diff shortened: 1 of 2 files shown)


aalexjo:
This addresses https://github.com/mikefarah/yq/issues/2518.

mikefarah:
Thanks @aalexjo !

aalexjo:
No problem, it would be great to perform a new snap release.

PR #2521, opened 2025-11-25
Bumped to core24 and removed riscv64

aalexjo:
- riscv64 was timing out, and was removed from snap builds.
- Non-interactive flag added for remote builds
- bumped snap base to core 24


mikefarah:
Thanks @aalexjo  - lets give it a go :)

PR #2548, opened 2025-12-17
Unwrap scalars in shell output mode.

### LINKED ISSUE #2504: Add "env" output format
Add "env" output format suitable for usage with Kubernetes configmap/secret create --from-env-file.  This differs from "shell" output format in that values should not be quoted, otherwise they become part of the value.

This could also be accomplished by the "-r" unwrap scalars option working with the "shell" output format.

flintwinters:
Resolves #2504 Adds `-r` support for `-o shell` outputs

First PR.

### DIFF
--- a/pkg/yqlib/encoder_shellvariables_test.go
+++ b/pkg/yqlib/encoder_shellvariables_test.go
@@ -168,8 +168,3 @@ func TestShellVariablesEncoderUnwrapScalar(t *testing.T) {
 	assertEncodesToUnwrapped(t, "c: true", "c=true")
 	assertEncodesToUnwrapped(t, "d: value with spaces", "d=value with spaces")
 }
-
-func TestShellVariablesEncoderDefaultQuoting(t *testing.T) {
-	assertEncodesTo(t, "a: Lewis Carroll", "a='Lewis Carroll'")
-	assertEncodesTo(t, "b: 123 456", "b='123 456'")
-}


pkg/yqlib/encoder_shellvariables.go
```diff
@@ -57,7 +57,13 @@ func (pe *shellVariablesEncoder) doEncode(w *io.Writer, node *CandidateNode, pat
 			// let's just pick a fallback key to use if we are encoding a single scalar
 			nonemptyPath = "value"
 		}
-		_, err := io.WriteString(*w, nonemptyPath+"="+quoteValue(node.Value)+"\n")
+		var valueString string
+		if pe.prefs.UnwrapScalar {
+			valueString = node.Value
+		} else {
+			valueString = quoteValue(node.Value)
```
mikefarah, line 64: Actually, looking at the quoteValue function, it only quotes if it determines that it needs to (non alpha numeric characters). 

I think with this change yq could emit invalid shell script...

flintwinters: Hi Mike,
Thanks for your reply.

The original, unchanged code used `quoteValue` indiscriminately.
Is this an issue with my PR or with the original issue?
If the former, I am not understanding the intended behavior.

Cheers.

mikefarah: looking at what quoteValue does; it only quotes if necessary :/ 

I'm now wondering if this is the right approach 🤔  There's also a "properties" file format that may be more suited - can you give a couple of examples that you need for your usecase?

flintwinters: Hi,
I am not the original opener of the issue, I was just implementing the change based on my understanding of it.

I would also be enlightened by some examples :sweat_smile: 
Maybe the issue should be closed if no examples are provided.

Best.

kknicker:
The "-r" Unwrap Scalars should only be used with "-o shell" if the output was intended to be used as an Env file per issue 2504.  The "-o prop" output format can't be used because properties file keys can contain characters which are invalid for Env files.

kknicker:
Clarification, the particular Env file format which issue 2504 was trying to address, was intended to be used with Kubernetes configmap/secret create --from-env-file.  All shell variable names are also valid Kubernetes configmap/secret keys.  But all properties file names are not.

mikefarah:
Alirght I think it all makes sense then 🤔  - @flintwinters can you rebase cause atm you've got commits in this PR that are already in master

flintwinters:
Hello all,
I believe this should have fixed the issue.  Thank you for your patience.
Regards.

PR #2553, opened 2025-12-20
Support negative parent indices

mikefarah:


mikefarah:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/mikefarah/yq/pull/2553#pullrequestreview-3600698235)

pkg/yqlib/operator_parent.go
```diff
@@ -35,9 +35,28 @@ func getParentOperator(_ *dataTreeNavigator, context Context, expressionNode *Ex
...
+		if prefs.Level < 0 {
+			// Count all parents
+			totalParents := 0
+			temp := candidate.Parent
+			for temp != nil {
+				totalParents++
+				temp = temp.Parent
+			}
+			// Convert negative index to positive
+			// -1 means last parent (root), -2 means second to last, etc.
```
ccoVeille, line 50: -2 the option is interesting, it could be documented maybe. I feel like you only documented the -1, right ?

I might be wrong

pkg/yqlib/doc/operators/parent.md
```diff
@@ -79,6 +79,46 @@ will output
...
+then
+```bash
+yq '.a.b.c | parent(-1)' sample.yml
+```
+will output
+```yaml
+a:
+  b:
+    c: cat
+```
```
ccoVeille, line 100: See https://github.com/mikefarah/yq/pull/2553#discussion_r2636938666

Maybe this


````suggestion
## Get the top (root) parent
Use negative numbers to get the top parents

Given a sample.yml file of:
```yaml
a:
  b:
    c:
      d: cat
```
then
```bash
yq '.a.b.c | parent(-2)' sample.yml
```
will output
```yaml
b:
  c:
    d: cat
```
````

I might be wrong in what -2 would bring, but the usage of `parent` with negative number would be clearer

This way `root` usage is clearer

PR #2557, opened 2025-12-20
Switch to uk

mikefarah:


.golangci.yml
```diff
@@ -14,6 +14,13 @@ linters:
     - unconvert
     - unparam
   settings:
+    misspell:
+      locale: UK
+      ignore-rules:
+        - color
+        - colors
+        - COLOR
```
ccoVeille: Misspell is not case sensitive unless I'm wrong

```suggestion
```

pkg/yqlib/encoder_hcl.go
```diff
@@ -173,19 +173,19 @@ func (he *hclEncoder) injectComments(output []byte, commentMap map[string]string
 	return []byte(result)
 }
 
-// colorizeHcl applies syntax highlighting to HCL output using fatih/color
+// colorizeHcl applies syntax highlighting to HCL output using fatih/color //nolint:misspell
```
ccoVeille: nolint is not working like this, I think.

`//nolint` is a Go directive. It should be on its own line. Without a space.

Also, using this may end up to have the nolint comment visible in the public godoc. Here you are safe as this is an unexported method.

I'm unsure what word is causing a problem here. the method godoc with `colorize`, `fatih/color` ?

Or some words in the method ? Because a //nolint:misspell in the godoc of method disables the role in the whole method.

I would try to remove it, then temporarily remove the exception you added to the golangci-lint and add them back progressively.

My guess is that you might not need it here

ccoVeille: I tend to configure golangci-lint `nolintlint` with require-explanation and require-specific to make sure I don't use nolint without a clear comment of what it is about.

https://golangci-lint.run/docs/linters/configuration/#nolintlint

.golangci.yml
```diff
@@ -14,6 +14,13 @@ linters:
...
+        - github.com/fatih/color
```
ccoVeille: I don't think misspell would use it. It expects single words

ccoVeille:
You may want to try this locally to see what else could have left behind in Markdown files, and maybe Go files

https://github.com/crate-ci/typos

```console
typos --locale en-gb
```

There is also a GitHub action is you are interested

EDIT: I didn't see you were using cspell https://github.com/mikefarah/yq/pull/2557#discussion_r2636925440

scripts/spelling.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-npx cspell --no-progress "**/*.{sh,go}"
\ No newline at end of file
+npx cspell --no-progress "**/*.{sh,go,md}"
```
ccoVeille, line 3: Oh I had no idea you were using cspell.

Then you could simply use this --locale en-GB here

PR #2560, opened 2025-12-24
feat: K8S KYAML output format support

robbat2:
Reference: https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/5295-kyaml/README.md
Co-authored-by: Codex <codex@openai.com>
Generated-with: OpenAI Codex CLI (partial)
Signed-off-by: Robin H. Johnson <rjohnson@coreweave.com>

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -292,3 +292,4 @@ nokyaml
 buildvcs
 behaviour
 GOFLAGS
+gocache
--- a/scripts/secure.sh
+++ b/scripts/secure.sh
@@ -3,9 +3,11 @@
 set -o errexit
 set -o pipefail
 
-if command -v gosec &> /dev/null
-then
-    gosec "${PWD}" ./...
-else
-  ./bin/gosec "${PWD}" ./...
-fi
+OPTS=(
+  -exclude-dir=vendor
+  -exclude-dir=.gomodcache
+  -exclude-dir=.gocache
+)
+
+command -v gosec &> /dev/null && BIN=gosec || BIN=./bin/gosec
+"${BIN}" "${OPTS[@]}" "${PWD}" ./...


project-words.txt
```diff
@@ -284,4 +284,7 @@ RDBMS
 expeñded
 bananabananabananabanana
 edwinjhlee
-flox
\ No newline at end of file
+flox
+unlabeled
```
mikefarah: unlabelled (double l) for UK english, which this project's locale

pkg/yqlib/kyaml_test.go
```diff
@@ -0,0 +1,107 @@
...
+				"    \"d\", # d line\n" +
+				"    \"e\",\n" +
+				"  ],\n" +
+				"  # trailing\n" +
+				"}\n",
+		},
+	}
+
+	for _, s := range scenarios {
+		testKYamlScenario(t, s)
```
mikefarah: Add 'documentScenarios' here so we can generate docs based off these tests :)

Makefile
```diff
@@ -53,7 +53,7 @@ tmp/dev_image_id: Dockerfile.dev scripts/devtools.sh
 build: build/dev
 
 .PHONY: build/dev
-build/dev: test *.go
+build/dev: *.go
```
mikefarah: this accidental? pls revert

pkg/yqlib/encoder_kyaml.go
```diff
@@ -0,0 +1,364 @@
...
+func (ke *kyamlEncoder) PrintDocumentSeparator(writer io.Writer) error {
+	if ke.prefs.PrintDocSeparators {
+		if err := writeString(writer, "---\n"); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (ke *kyamlEncoder) PrintLeadingContent(writer io.Writer, content string) error {
```
mikefarah, line 30: is this the same logic as the yaml encoder? Could it be pulled out into a common function?

pkg/yqlib/kyaml_test.go
```diff
@@ -0,0 +1,107 @@
...
+				"  # head b\n" +
+				"  b: 2,\n" +
+				"  c: [\n" +
+				"    # head d\n" +
+				"    \"d\", # d line\n" +
+				"    \"e\",\n" +
+				"  ],\n" +
+				"  # trailing\n" +
+				"}\n",
+		},
```
mikefarah, line 485: good to add some more tests in here, plain scalars, maybe converting from yaml to kyaml to show the difference (particularly when you add documentScenarios in.

pkg/yqlib/kyaml_test.go
mikefarah: You'll also need to add a header/kyaml.md (see the other formats for examples 👍🏼 )

mikefarah:
Nice work codex + @robbat2  - couple of comments to resolve

mikefarah:
Oh other thing to add are the compiler directives to allow for minimal builds (e..g `//go:build !yq_nohcl`)

Makefile
robbat2: will revert for now; but I was finding it hindered rapid development. I do want to test - but the slow test targets in the Makefile shouldn't be a pre-req for a quick development build.

project-words.txt
robbat2: fixed

pkg/yqlib/kyaml_test.go
robbat2: done

pkg/yqlib/encoder_kyaml.go
robbat2: done

pkg/yqlib/kyaml_test.go
robbat2: done; and I told `agents.md` to do this in future.

robbat2:
@mikefarah
The KYAML part is done to your additional requests; codex handled it nicely when I added the `documentScenarios` hint to `agents.md`

 i'm thinking these build improvements (they didn't come from Codex) should maybe be a seperate PR. Esp the gosec change that gives a 3+ minute speedup.

There's one that still annoying, but I think it's a bug in `golanglint-ci` - it's ignoring the `buildvcs=false`; and then failing when I run `make check`

PR #2569, opened 2026-01-10
feat: JSON5 support

robbat2:
Reference: https://spec.json5.org/
Co-authored-by: Codex <codex@openai.com>
Generated-with: OpenAI Codex CLI (partial)
Signed-off-by: Robin H. Johnson <rjohnson@coreweave.com>

### DIFF
--- a/acceptance_tests/inputs-format-auto.sh
+++ b/acceptance_tests/inputs-format-auto.sh
@@ -5,6 +5,7 @@ setUp() {
   rm test*.toml 2>/dev/null || true
   rm test*.tfstate 2>/dev/null || true
   rm test*.json 2>/dev/null || true
+  rm test*.json5 2>/dev/null || true
   rm test*.properties 2>/dev/null || true
   rm test*.csv 2>/dev/null || true
   rm test*.tsv 2>/dev/null || true
@@ -28,6 +29,29 @@ EOM
   assertEquals "$expected" "$X"
 
   X=$(./yq ea test.json)
+  assertEquals "$expected" "$X"
+}
+
+testInputJson5() {
+  cat >test.json5 <<'EOL'
+{ /* hello */
+  mike: { things: "cool", },
+}
+EOL
+
+  read -r -d '' expected << EOM
+// hello
+{
+  "mike": {
+    "things": "cool"
+  }
+}
+EOM
+
+  X=$(./yq test.json5)
+  assertEquals "$expected" "$X"
+
+  X=$(./yq ea test.json5)
   assertEquals "$expected" "$X"
 }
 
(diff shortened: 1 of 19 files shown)


robbat2:
new push has more test coverage for cases that I didn't think of in the initial write

pkg/yqlib/encoder_json5.go
```diff
@@ -0,0 +1,347 @@
...
+		}
+		return "false"
+	case "!!int":
+		value, err := node.GetValueRep()
+		if err != nil {
+			return "null"
+		}
+		intBytes, err := json.Marshal(value)
+		if err != nil {
+			return "null"
```
mikefarah, line 253: should log warning with error details here

pkg/yqlib/encoder_json5.go
```diff
@@ -0,0 +1,347 @@
...
+		}
+		return string(intBytes)
+	case "!!float":
+		return json5FloatString(node.Value)
+	case "!!str":
+		stringBytes, err := json.Marshal(node.Value)
+		if err != nil {
+			return "null"
```
mikefarah, line 261: same here

pkg/yqlib/encoder_json5.go
```diff
@@ -0,0 +1,347 @@
...
+
+	floatBytes, err := json.Marshal(parsed)
+	if err != nil {
+		return "null"
+	}
+	return string(floatBytes)
+}
+
+func writeJSON5CommentBlock(writer io.Writer, comment string, indentString string, depth int) error {
+	comment = strings.ReplaceAll(comment, "\r\n", "\n")
```
mikefarah, line 306: I recently had to do a bunch of fixes in the yaml decoder (https://github.com/mikefarah/yq/blob/542801926fe22c8d21e0698f5b9d952301fe636f/pkg/yqlib/decoder_yaml.go#L57) to support both line endings 😅  - would be good to have this also support both to avoid a similar bug being raised

pkg/yqlib/encoder_json5.go
```diff
@@ -0,0 +1,347 @@
...
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue
+		}
+		line = strings.TrimSpace(strings.TrimPrefix(line, "#"))
+		line = strings.TrimSpace(strings.TrimPrefix(line, "//"))
+		if err := writeString(writer, indent); err != nil {
+			return err
+		}
+		if err := writeString(writer, "// "+line+"\n"); err != nil {
```
mikefarah, line 324: not sure why you're not keeping the original comment delimiter?

acceptance_tests/inputs-format-auto.sh
```diff
@@ -31,6 +32,29 @@ EOM
...
 
+testInputJson5() {
+  cat >test.json5 <<'EOL'
+{ /* hello */
+  mike: { things: "cool", },
+}
+EOL
+
+  read -r -d '' expected << EOM
+// hello
```
mikefarah, line 43: would be good to round trip the delimiter, this will definitely be raised as an issue

PR #2583, opened 2026-01-26
feat: Add --yaml-compact-seq-indent / -c flag for compact sequence indentation

### LINKED ISSUE #1841: Allow lists to not be indented

I wish I could force my YAML output to have a list indentation style of :

```yaml
key:
  value
list:
- one
- two
- three
```

instead of:

```yaml
key:
  value
list:
  - one
  - two
  - three
```

While I'm new to yq, I don't think this is currently possible after reading https://mikefarah.gitbook.io/yq/usage/output-format#indent , skimming the man page, and studying `--help`

> Note that lists are indented at the same level as the map key at indent level 2, but are more deeply indented at indent level 4 and greater. This is (currently) a quirk of the underlying yaml parser

However, I'd be super happy to be proven wrong!

Not sure, maybe another CLI flag  that pertains to lists or a `style` operator change?

Maybe I could write some sort of shell script to post-process my YAML file?

**Additional context**

I want to use yq to format my large [octodns](https://github.com/octodns/octodns) YAML files by top-level key, but keep the pre-existing comments and list indentation. Keeping the comments already seems to work, so that just leaves the list indentation.

I totally understand if this CLI is limited by the underlying YAML lib and so couldn't do this without an unreasonable amount of work, but I didn't see any issues with a [search](https://github.com/mikefarah/yq/issues?q=dis%3Aissue+list+indented), so thought I'd ask.

Thanks so much for your work on `yq` and releasing it to the public!

jfenal:
## Summary

- Adds a new CLI flag `--yaml-compact-seq-indent` (shortcut `-c`) that enables compact sequence indentation
- When enabled, `'- '` is considered part of the indentation, resulting in sequences that align with their parent key
- Leverages the `CompactSeqIndent()` method from the underlying `go.yaml.in/yaml/v4` library

## Example

**Default behavior:**
```yaml
parent:
  items:
    - one
    - two
```

**With `--yaml-compact-seq-indent` / `-c`:**
```yaml
parent:
  items:
  - one
  - two
```

## Test plan

- [x] Verified build compiles successfully
- [x] Tested flag works with both long and short forms
- [x] Verified default behavior is unchanged (sequences remain indented)
- [ ] Run full test suite

Closes #1841

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -203,6 +203,7 @@ yq -P -oy sample.json
 	}
 	rootCmd.PersistentFlags().BoolVarP(&yqlib.ConfiguredYamlPreferences.LeadingContentPreProcessing, "header-preprocess", "", true, "Slurp any header comments and separators before processing expression.")
 	rootCmd.PersistentFlags().BoolVarP(&yqlib.ConfiguredYamlPreferences.FixMergeAnchorToSpec, "yaml-fix-merge-anchor-to-spec", "", false, "Fix merge anchor to match YAML spec. Will default to true in late 2025")
+	rootCmd.PersistentFlags().BoolVarP(&yqlib.ConfiguredYamlPreferences.CompactSequenceIndent, "yaml-compact-seq-indent", "c", false, "Use compact sequence indentation where '- ' is considered part of the indentation.")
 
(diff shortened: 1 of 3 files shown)


jfenal:
Full disclosure: this pull request was vibe coded with Cursor.

PR #2585, opened 2026-01-28
docs: add hints for common syntax edge cases

chrishamant:
Adds informational hints to operator documentation covering behaviors
  that may differ from user expectations, particularly for those
  familiar with jq:

  - slice-array: clarify that slicing only works on arrays, not strings
  - select: document behavior when piping select to a literal value
  - equals: note bash history expansion with != operator
  - create-collect-into-object: mention that shorthand {key} syntax is not supported
  - collect-into-array: explain difference between [.x | .[]] and [.x[]]

This came from dorking around with AI agents trying to get them to use utilize `yq` correctly. It seems that `jq` takes up quite a bit of latent space and they (at least Claude + ChatGPT) seem to assume that "something that works in `jq` should work in `yq`". I'm not not a big MCP fan atm and have been relying on building a quality "skill" that cribs patterns iteratively and I kept getting output that would consistently assume incorrect usages... I've honed this skill a bit now reifying my learned/observed stuff with this `gitbook` as source of truth. Thought I might be able to contribute back slightly by asking "what could we improve in the docs" vs just going the other way. 

I know unprompted AI slop is the the bane of everyone's existence right now and I'm trying not to pile that on to open source maintainers... I wanted to offer up this PR in which I've tried to be somewhat surgical and match tone where I can documenting some of the edge cases/assumptions I've seen my agents make in trying to use your great tool / competing for latent space with `jq`. I'm 97% sure this is hallucination free as it is validated by a variety of tests/scripts I built while sorting out issues with my skill watching agents fail to use `yq`.  I'm not at all married to this won't be offended if you just want to close or crib parts... (the skill I have now works pretty well for me)

Thanks again for your time / `yq`!

### DIFF
--- a/operators/collect-into-array.md
+++ b/operators/collect-into-array.md
@@ -2,6 +2,19 @@
 
 This creates an array using the expression between the square brackets.
 
+{% hint style="warning" %}
+
+_Note_ the placement of `|` when collecting. These two forms behave differently:
+
+```bash
+# Pipe then splat - creates separate context
+[.items | .[] | has("id")]
+
+# Splat directly on path - more common pattern
+[.items[] | has("id")]
+```
+
+{% endhint %}
 
 ## Collect empty
 Running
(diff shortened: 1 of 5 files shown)


mikefarah:
Thanks @chrishamant  - these docs are generated from the tests though 😅  -

chrishamant:
Ah no worries! My bad

Issue #2588, opened 2026-01-31
TOML: comments inside table cause hierarchy to be flattened


tomers:
### Summary

When parsing a TOML file that has **comments inside a table section**, yq flattens that table's keys to the document root. Without comments, the hierarchy is preserved correctly. This makes `.toolchain.channel` return `null` (and `.channel` return the value) when comments are present, and `.toolchain.channel` work correctly when comments are removed.

### Environment

- **yq version:** v4.52.1 (mikefarah/yq)
- **OS:** Linux

### Steps to reproduce

**File with comments (flattened — bug):**

Save as `rust-toolchain-with-comments.toml`:

```toml
[toolchain]
# Use stable to match production Docker builds
# To pin a specific version: channel = "1.92.0"
channel = "stable"
components = ["llvm-tools-preview"]
```

```bash
yq .toolchain.channel rust-toolchain-with-comments.toml
# Actual: null
# Expected: stable

yq .channel rust-toolchain-with-comments.toml
# Actual: stable
# Expected: null (channel is under [toolchain], not at root)
```

**Same file without comments (correct):**

Save as `rust-toolchain-no-comments.toml`:

```toml
[toolchain]
channel = "stable"
components = ["llvm-tools-preview"]
```

```bash
yq .toolchain.channel rust-toolchain-no-comments.toml
# Actual: stable (correct)

yq .channel rust-toolchain-no-comments.toml
# Actual: null (correct)
```

### Expected behavior

Comments inside a TOML table should not change the structure. In both files, `channel` and `components` are keys of the `[toolchain]` table, so:

- `.toolchain.channel` should be `stable`
- `.channel` should be `null` (no root-level `channel`)

### Actual behavior

With comments present between `[toolchain]` and the key-value pairs, yq puts `channel` and `components` at the document root instead of under `toolchain`, so `.toolchain.channel` is `null` and `.channel` is `stable`.

### Possible cause

The TOML decoder (or its integration in yq) may be associating key-value pairs that follow comments with the root context instead of the current table context, or comment handling is altering the table scope.

mikefarah:
Fixed in 4.52.2

PR #2589, opened 2026-01-31
Fix TOML table scope after comments

tomers:
## Summary
- Fix TOML decode so standalone comments immediately inside a table/array-table don’t terminate the table scope.
- Prevents keys following those comments from being incorrectly assigned at the document root (issue #2588).
- Adds regression test covering the scenario.

## Test plan
- `go test ./...`


tomers:
Observed Behavior: Regression.
Last Known Working Version: v4.50.1
Broken In: v4.52.1

mikefarah:
Thanks for the fix @tomers ! Fwiw - next time run `make local build` to run all the checks :)

tomers:
> next time run make local build to run all the checks
Sure! It would be best if you make it run automatically when a PR is created (reference: `esphome/esphome`)

Issue #2587, opened 2026-01-31
go install command not working for latest releases


tkrop:
**Describe the bug**

When trying to install the latest `yq` version via 

```bash
go install -buildvcs=auto github.com/mikefarah/yq/v4@latest
```

I get the following error message:

```txt
go: downloading github.com/mikefarah/yq/v4 v4.52.1
go: github.com/mikefarah/yq/v4@latest:
create zip: .github/instructions/*.instructions.md:
malformed file path ".github/instructions/*.instructions.md": invalid char '*'
```

Also trying to fallback to a previous version is leading to an error due to deprecating the previous versions, e.g.:

```txt
go: github.com/mikefarah/yq/v4@v4.48.2:
loading deprecation for github.com/mikefarah/yq/v4: github.com/mikefarah/yq/v4@v4.52.1:
verifying go.mod: github.com/mikefarah/yq/v4@v4.52.1/go.mod:
reading https://sum.golang.org/lookup/github.com/mikefarah/yq/v4@v4.52.1: 404 Not Found
	server response: not found: create zip: .github/instructions/*.instructions.md:
	malformed file path ".github/instructions/*.instructions.md": invalid char '*'
```

The file naming with `*` seems to violate the `zip`-tool dependencies used by `go install`.

Moulick:
Yep, This is breaking CI in several repos

theyoprst:
`go install` doesn't work for any release of yq, not just the latest. Because go install still tries to access the latest release and fails:

```
➜ go install github.com/mikefarah/yq/v4@v4.44.3
go: github.com/mikefarah/yq/v4@v4.44.3: loading deprecation for github.com/mikefarah/yq/v4: github.com/mikefarah/yq/v4@v4.52.1: verifying go.mod: github.com/mikefarah/yq/v4@v4.52.1/go.mod: reading https://sum.golang.org/lookup/github.com/mikefarah/yq/v4@v4.52.1: 404 Not Found
        server response: not found: create zip: .github/instructions/*.instructions.md: malformed file path ".github/instructions/*.instructions.md": invalid char '*'

➜ go install github.com/mikefarah/yq/v4@v4.50.1
go: downloading github.com/mikefarah/yq/v4 v4.50.1
[...]

PR #2590, opened 2026-01-31
Rename instructions file to fix go install failure

theyoprst:
The asterisk character in *.instructions.md caused go install to fail with 'malformed file path' error. Renamed to bug-fix.instructions.md to resolve issue #2587.


mikefarah:
Thanks for raising this, ended up just calling it instructions.md - not sure how I checked it in the '*.instruction' 🤦🏼

Issue #2592, opened 2026-02-02
Recent update broke comments in arrays in TOML


Artemigos:
**Describe the bug**
I had a config TOML file with comments in arrays like this:

```toml
[section]
the_array = [
  # comment
  "value 1",

  # comment
  "value 2",
]
```

This broke when running `yq . -oj ~/tmp/yq-bug.toml` and is now reporting:

```
Error: bad file '[file path here]': unsupported type Comment
```

Running without `-oj` has the same problem.

Version of yq: 4.52.2
Operating system: linux
Installed via: binary release

**Additional context**
I wasn't sure if this syntax is even supported, but according to spec it is: https://toml.io/en/v1.1.0#:~:text=Any%20number%20of%20newlines%20and%20comments%20may%20precede%20values%2C%20commas%2C%20and%20the%20closing%20bracket%2E

Galdanwing:
I'm experiencing the same issue with `yq` parsing a `pyproject.toml` file in my case. Interestingly, it's occurring across multiple versions:
- Pinned version in GitHub CI: `yq@v4.34.1` (3 years old)
- Locally installed via Brew: `v4.52.2`

Since both versions started failing today, this to me suggests an unpinned dependency was updated and introduced a breaking change.

benjamin-pilgrim:
Also seeing this issue. The yq github action doesn't pin the tagged version of the docker image, just `docker://mikefarah/yq:4-githubaction`.

flying-sheep:
So this will be fixed by #2595?

mikefarah:
Fixed in 4.52.4 :)

PR #2595, opened 2026-02-03
Fixing comments in TOML arrays #2592

mikefarah:


asd585140-glitch (commented):
[SupportedOperations] // no mathematical operators should be generated
public readonly partial struct CustomerId : IDomainValue<int>
{
	public static void Validate(int value)
	{
		if (value <= 0)
			throw new InvalidDomainValueException("Value must be a positive number");
	}

	public static int Default => 1;
}

public sealed class Transaction
{
	public CustomerId FromId { get; set; }
	public CustomerId? ToId { get; set; }
	public PositiveAmount Amount { get; set; }
	public PositiveAmount? Fees { get; set; }
}

public static void JsonSerializationAndDeserialization()
{
	var amount = new Transaction()
        {
            Amount = 100.523m,
            Fees = null,
            FromId = 1,
            ToId = null
        };

    var jsonValue = JsonSerializer.Serialize(amount); //this will produce the same result as changing customerId to int and PositiveAmount to decimal
    var newValue = JsonSerializer.Deserialize<Transaction>(jsonValue)
}

asd585140-glitch:
var jsonValue = JsonSerializer.Serialize(amount); //this will produce the same result as changing customerId to int and PositiveAmount to decimal
var newValue = JsonSerializer.Deserialize<Transaction>(jsonValue)

PR #2610, opened 2026-02-23
docs: resolve latest version dynamically in wget install example

gitmpr:
## Summary

- Replaces the hardcoded `VERSION=v4.2.0` in the wget install example with a dynamic lookup via the GitHub API
- Moves the inline comment to the `PLATFORM` variable, which is the only thing users actually need to adjust

## Before

```bash
# Set your platform variables (adjust as needed)
VERSION=v4.2.0
PLATFORM=linux_amd64
```

## After

```bash
PLATFORM=linux_amd64  # adjust as needed
VERSION=$(wget -qO- https://api.github.com/repos/mikefarah/yq/releases/latest | jq -r .tag_name)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,9 +82,8 @@ Use wget to download pre-compiled binaries. Choose your platform and architectur
 
 **For Linux (example):**
 ```bash
-# Set your platform variables (adjust as needed)
-VERSION=v4.2.0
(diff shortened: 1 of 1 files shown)


mikefarah:
Hmm but then to follow these instructions you'd first need to install jq...

Issue #515, opened 2020-08-13
yq write strips completely blank lines from the output


scanfield:
**Is your feature request related to a problem? Please describe.**
```
foo:
  bar: 1

  baz: 2
```

when run through `yq w - foo.baz 3`

produces

```
foo:
  bar: 1
  baz: 3
```

**Describe the solution you'd like**
Keep my extra blank line (it's better for readability / produces less of a diff)

warder:
Same story, sorry but this issue looks more like a bug nether then enhancement.
When you process yaml file with yq it corrupts a whole file

AceHack:
Any update on this, this is a really nice feature.

sathiyams:
Any update on this ? It's getting difficult when it comes to readability

mikefarah:
This is an effect of the underlying yaml parser, an issue was raised there https://github.com/go-yaml/yaml/issues/627 - the owner said 

```
..the content when re-encoded will not
 have its original textual representation preserved. An effort is made to
 render the data plesantly, and to preserve comments near the data they
 describe, though. 
```

arcesino:
I've been dealing with this issue for a couple of days when updating very large YAML files and found a workaround using `diff` & `patch` commands that restores the stripped blank lines in most of the cases. Suppose you have the following YAML file:

```yaml
doc:
  version: 1.0.0
  name: numbers & letters

numbers:
  - 1

letters:
  - a
```
we call this file `a.yaml`. Now let's update the version using `yq` and store the result in new file `a-updated.yaml`:

```sh-session
yq e '.doc.version = "1.0.1"' a.yaml > a-updated.yaml
```
as expected, command above stripped all blank lines so `a-updated.yaml` looks like:
```yaml
doc:
  version: 1.0.1
  name: numbers & letters
numbers:
  - 1
letters:
  - a
```
at this point, the first step to get the blank lines back is to create a diff file that ignores blank lines changes:
```sh-session
diff -U0 -w -b --ignore-blank-lines a.yaml a-updated.yaml > a.diff
```
`a.diff` looks like this:
```diff
--- a.yaml	2021-04-30 15:28:38.000000000 -0500
+++ a-updated.yaml	2021-04-30 15:18:53.000000000 -0500
@@ -2 +2 @@
-  version: 1.0.0
+  version: 1.0.1
``` 
then final step is to patch original file with the diff:
```sh-session
patch a.yaml < a.diff
```
after that, the original file looks like:
```yaml
doc:
[...]

PR #2619, opened 2026-03-04
Preserve empty lines in HCL

jtyr:
This PR improves HCL roundtrip fidelity by preserving blank lines between blocks and attributes, and correctly re-encoding raw HCL expressions inside flow-style object attributes without wrapping them in quotes.

### DIFF
--- a/pkg/yqlib/candidate_node.go
+++ b/pkg/yqlib/candidate_node.go
@@ -100,6 +100,9 @@ type CandidateNode struct {
 	// For formats like HCL and TOML: indicates that child entries should be emitted as separate blocks/tables
 	// rather than consolidated into nested mappings (default behaviour)
 	EncodeSeparate bool
+	// For formats like HCL: indicates that a blank line preceded this node in the original source,
+	// so the encoder should emit a blank line before it to preserve formatting.
+	BlankLineBefore bool
 }
 
 func (n *CandidateNode) CreateChild() *CandidateNode {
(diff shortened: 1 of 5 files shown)


pkg/yqlib/candidate_node.go
```diff
@@ -411,7 +414,8 @@ func (n *CandidateNode) doCopy(cloneContent bool) *CandidateNode {
 		EvaluateTogether: n.EvaluateTogether,
 		IsMapKey:         n.IsMapKey,
 
-		EncodeSeparate: n.EncodeSeparate,
+		EncodeSeparate:  n.EncodeSeparate,
+		BlankLineBefore: n.BlankLineBefore,
```
mikefarah, line 418: Why not put blank lines in the HeaderComment?

jtyr: HeadComment is meant to hold comment text. Overloading it with formatting whitespace (e.g. a leading `\n` for blank-line-only nodes, or `\n# actual comment` when both are present) mixes two different concerns. Nodes can have a blank line without any comment, and vice versa.

There's also a processing order issue. In the encoder, comments are injected first, then blank lines in a second pass. This makes sure blank lines end up before the comment+attribute group, not between them. Merging both into `HeadComment` would require a single pass to handle this correctly, making the comment injection logic more   complex.

The main reason though is cross-format extensibility. The `BlankLineBefore` field on `CandidateNode` is format-agnostic. Right now only HCL uses it, but the same mechanism could be reused by other format decoders (TOML, Properties, potentially YAML) without each one having to encode blank line info into comment strings and then parse it back out. Each decoder just sets a boolean, and each encoder decides how to act on it.

mikefarah: Hmm have to think about this one a little more. There's been a separate, long discussion on blank lines here: https://github.com/mikefarah/yq/issues/515

I don't see how this would solve the blank line issue properly. Can't capture multiple blank lines, are the blank lines before or after the head comment...

PR #2644, opened 2026-04-06
Fix panic and OOM in repeatString for large repeat counts

jandubois:
The existing check (count > 10 million) does not account for string length. A 68-byte string repeated 35 trillion times passes the count check but panics in strings.Repeat with "makeslice: len out of range". Smaller counts (e.g. 10 million * 6-byte string = 60 MB) cause OOM on memory-constrained environments like OSS-Fuzz (2560 MB limit).

Replace the count-only check with a result size check: the product of string length and repeat count must not exceed 10 MiB. Use division (len > limit/count) instead of multiplication (len*count > limit) to avoid integer overflow — a large count can wrap the product to a negative value, bypassing the guard entirely.

Fixes at least four OSS-Fuzz bugs found via Lima's FuzzEvaluateExpression:
  https://issues.oss-fuzz.com/issues/418818862 (makeslice overflow)
  https://issues.oss-fuzz.com/issues/422001683 (timeout from huge alloc)
  https://issues.oss-fuzz.com/issues/383195001 (OOM, 3 GB allocation)
  https://issues.oss-fuzz.com/issues/385180606 (OOM, 97 TB allocation)

### DIFF
--- a/pkg/yqlib/operator_multiply.go
+++ b/pkg/yqlib/operator_multiply.go
@@ -155,8 +155,10 @@ func repeatString(lhs *CandidateNode, rhs *CandidateNode) (*CandidateNode, error
 		return nil, err
 	} else if count < 0 {
 		return nil, fmt.Errorf("cannot repeat string by a negative number (%v)", count)
-	} else if count > 10000000 {
-		return nil, fmt.Errorf("cannot repeat string by more than 100 million (%v)", count)
+	}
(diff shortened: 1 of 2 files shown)


WhyNotHugo:
Tests fail on all 32-bit platforms.

jandubois:
I believe I have fixed this in #2680, but have no way to verify it on an actual 32-bit system.

PR #2656, opened 2026-04-11
Quote top-level string scalars whose bare form re-parses as a container

### LINKED ISSUE #2608: A single string scalar as output is not quoted if necessary (breaking roundtrip safety)
**Describe the bug**

If the output from yq is a single string, then it will not be quoted properly in all cases.

If the string's content is the valid YAML encoding of something else (e.g. an integer, a map or a sequence), the output YAML will not encode that string, but the data encoded by that string when interpreted as a YAML value.

In other words: yq fails to detect if encoding a single string scalar as output, whether this string needs to be quoted. You could also say it treats a string scalar output as if you have given the `-r` flag in all cases.

Version of yq: 4.53.2
Operating system: linux
Installed via: docker

**Input Yaml**
data1.yml:
```yaml
"this: should really work"
```

**Command**
The command you ran:
```
yq eval . data1.yaml
```

**Actual behaviour**

```yaml
this: should really work
```

This should only happen, if we called `yq` with the `-r` option to enable raw output of scalars.

**Expected behaviour**

```yaml
"this: should really work"
```

**Additional context**

The problem does not depend on the input being top-level, but only on the output being a string scalar. It occurs the same for:

data2.yaml
```yaml
foo: |
  a: a
  b: b
```
```
yq .foo data2.yaml
```

Which yields the output
```yaml
a: a
b: b
```

It also happens if the output is a stream documents (if one of the documents is just a string scalar).

I could not reproduce this behaviour for other cases (such as when returning a sequence or map) and I haven't looked at the code yet to see whether this could be triggered in other cases.

barry3406:
The yaml encoder has a fast-path for unwrapping top-level scalars: when `UnwrapScalar` is set (the default for yaml output) it writes `node.Value` verbatim and skips the regular encoder. The fast-path never checks whether the bare form round-trips, so any `!!str` whose value happens to be a valid YAML mapping or sequence is silently reinterpreted on the next read.

Reproducer:

```
printf '"this: should really work"\n' | yq -p yaml -o yaml
```

Before this fix the second invocation parses the output as a single-key map, destroying the original string; the reporter's second example (`yq .foo` on a `foo: |` block containing `a: a` / `b: b`) hits the same path through the literal style.

Guard the fast-path with `bareStringNeedsQuoting`, which re-parses `node.Value` with yaml.v4 and reports whether the result is something other than a scalar. When it returns true the encoder falls through to the normal `encoder.Encode` path, which picks up `LiteralStyle` or `DoubleQuotedStyle` as appropriate. The check is deliberately narrow:

- only `!!str` nodes are examined, so existing tag tests that emit `!!int`/`!!str` as a bare scalar are untouched;
- only structural reinterpretations (map/sequence/alias) trigger the fallback, so plain strings that re-read as integers, booleans, or nulls keep their current unwrapped form;
- an unparseable value (e.g. one containing NUL) stays on the fast-path, so the NUL-aware printer in `printer.go` still sees the raw bytes.

The new table-driven test `TestYamlEncoderUnwrapScalarRoundtripSafety` covers the colon, dash, and multiline-map cases from the report and asserts that safe plain strings, digits, `null`, and `!!int` still emit bare. The `decode yaml document` base64 scenario previously asserted the buggy `a: apple\n` bare output; it now expects the block-literal form that round-trips back to the same string.

Fixes #2608

### DIFF
--- a/pkg/yqlib/base64_test.go
+++ b/pkg/yqlib/base64_test.go
@@ -81,10 +81,13 @@ var base64Scenarios = []formatScenario{
 		scenarioType: "decode",
 	},
 	{
-		skipDoc:      true,
-		description:  "decode yaml document",
-		input:        base64EncodedYaml,
-		expected:     base64DecodedYaml + "\n",
+		skipDoc:     true,
+		description: "decode yaml document",
+		input:       base64EncodedYaml,
+		// The decoded payload ("a: apple\n") would re-parse as a map if
+		// emitted bare, so the yaml encoder keeps it as a block literal to
+		// preserve roundtrip safety. See issue #2608.
+		expected:     "|\n    a: apple\n",
 		scenarioType: "decode",
 	},
 	{
--- a/pkg/yqlib/encoder_yaml.go
+++ b/pkg/yqlib/encoder_yaml.go
@@ -35,7 +35,7 @@ func (ye *yamlEncoder) Encode(writer io.Writer, node *CandidateNode) error {
 	if strings.Contains(node.LeadingContent, "\r\n") {
 		lineEnding = "\r\n"
 	}
-	if node.Kind == ScalarNode && ye.prefs.UnwrapScalar {
+	if node.Kind == ScalarNode && ye.prefs.UnwrapScalar && !bareStringNeedsQuoting(node) {
 		valueToPrint := node.Value
 		if node.LeadingContent == "" || valueToPrint != "" {
 			valueToPrint = valueToPrint + lineEnding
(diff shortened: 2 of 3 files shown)


pkg/yqlib/base64_test.go
```diff
@@ -81,10 +81,13 @@ var base64Scenarios = []formatScenario{
...
-		description:  "decode yaml document",
-		input:        base64EncodedYaml,
-		expected:     base64DecodedYaml + "\n",
+		skipDoc:     true,
+		description: "decode yaml document",
+		input:       base64EncodedYaml,
+		// The decoded payload ("a: apple\n") would re-parse as a map if
+		// emitted bare, so the yaml encoder keeps it as a block literal to
+		// preserve roundtrip safety. See issue #2608.
+		expected:     "|\n    a: apple\n",
```
mikefarah, line 90: This is what worries me:
a) how would you now emit this as a bare string if you wanted to; 
b) this change could break a whole bunch of scripts / CI/CD pipelines for people

barry3406: Yeah that's a valid concern. A couple thoughts:

(a) Users can still get the raw value with `-o=raw` / `--outputformat=raw`, which bypasses the yaml encoder entirely. So there is an escape hatch.

(b) On breakage — this only kicks in for strings whose bare form re-parses as a map or sequence (like `a: b` or `- item`). Plain strings, numbers, bools are untouched. So it's a pretty narrow change, but I get that even narrow changes can bite someone.

If you'd prefer a flag to opt into this behavior instead of making it the default, I can rework it that way.

PR #2680, opened 2026-04-23
Fix repeatString overflow test on 32-bit platforms

jandubois:
The test literal "ab" * 4611686018427387904 (2^62) exceeds MaxInt32, so parseInt rejects it before the size guard runs. Compute the count with 1 << (bits.UintSize - 2) to yield 2^30 on 32-bit and 2^62 on 64-bit. Both values, when doubled by len("ab"), wrap past MaxInt and bypass a naive len*count guard, exercising the division-safe check added in #2644.

Ref https://github.com/mikefarah/yq/pull/2644#issuecomment-4297857398

I've verified that the modified test case still triggers the original panic on 64-bit platforms unless the fix from #2644 is applied.

I've also verified that this change fixes cross-compilation to 32-bit mode.

I am **NOT** able to test the 32-bit binary itself, so I don't know if it will pass tests, and I don't know if this code will still trigger the original bug. From a pure math perspective it should.

### DIFF
--- a/pkg/yqlib/operator_multiply_test.go
+++ b/pkg/yqlib/operator_multiply_test.go
@@ -2,6 +2,7 @@ package yqlib
 
 import (
 	"fmt"
+	"math/bits"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
Thanks!

Builds fine on [armhf](https://gitlab.alpinelinux.org/alpine/aports/-/jobs/2318571) and [x86](https://gitlab.alpinelinux.org/alpine/aports/-/jobs/2318577)

> I am NOT able to test the 32-bit binary itself, so I don't know if it will pass tests, and I don't know if this code will still trigger the original bug. From a pure math perspective it should.

Probably just running on x86 on CI should suffice for detecting regressions, but this works as-is.

mikefarah (approved):

Issue #2677, opened 2026-04-21
<< short merge tags are not !!merge modified AND are now quoted, breaking some parsers


nathanmcgarvey-modopayments:
I'm not sure if this is a bug or just something I'd like search engines to find:

due to https://github.com/yaml/go-yaml/commit/fbd745be58df058ec8eb7d7b42468458f8d8ec64#r182994094
in combination with the version bump here in https://github.com/mikefarah/yq/commit/8c018da9c94077c03e124208b4e8d70ad6324856 from rc.3 to rc.4 of go-yaml.

Merge tags (<<) from YAML 1.1, when using operators such as `*+` seem to now only quote the << operator and !! is not added, either.

 ```shell
go run yq.go eval-all '. as $item ireduce ({}; . *+ $item)' < /tmp/sample.yaml
```

where sample.yaml is:
```yaml
---
defaults: &defaults
  foo: bar
  bat: baz
not_defaults:
  <<: *defaults
  foo: pizza
---
defaults2: &defaults2
  foo: bar
  bat: baz
not_defaults2:
  <<: *defaults2
  foo: pizza
```

example output from rc.3 of go-yaml which is used in v4.52.5 of yq:

```yaml
---
defaults: &defaults
  foo: bar
  bat: baz
not_defaults:
  <<: *defaults
  foo: pizza
defaults2: &defaults2
  foo: bar
  bat: baz
not_defaults2:
  <<: *defaults2
  foo: pizza
```

example output from rc.4 of go-yaml which is used in v4.53.2 of yq:

```yaml
---
defaults: &defaults
  foo: bar
  bat: baz
not_defaults:
  "<<": *defaults
  foo: pizza
defaults2: &defaults2
  foo: bar
  bat: baz
not_defaults2:
  "<<": *defaults2
  foo: pizza
```

nathanmcgarvey-modopayments:
If this is a bug, please mark it as such, if this is "just how it goes", then some documentation on alternatives may be nice. (E.g. manually using !!merge, if that's correct.)

W-Floyd:
I saw this a little bit ago too. It seems to apply unevenly to instances of this syntax.

Minimal example I'd made at the time:
```
package main

import (
	"os"

	"github.com/mikefarah/yq/v4/pkg/yqlib"
)

var yqExpr = `. as $item ireduce ({}; . *+ $item)`

func main() {

	data, _ := os.ReadFile("input.yaml")

[...]

PR #2705, opened 2026-05-12
Fix for #2677

W-Floyd:
Depends on https://github.com/yaml/go-yaml/pull/348

### DIFF
--- a/pkg/yqlib/candidate_node.go
+++ b/pkg/yqlib/candidate_node.go
@@ -46,7 +46,11 @@ const (
 func createStringScalarNode(stringValue string) *CandidateNode {
 	var node = &CandidateNode{Kind: ScalarNode}
 	node.Value = stringValue
-	node.Tag = "!!str"
+	if stringValue == "<<" {
+		node.Tag = "!!merge"
+	} else {
+		node.Tag = "!!str"
+	}
 	return node
 }
 
--- a/pkg/yqlib/operator_anchors_aliases_test.go
+++ b/pkg/yqlib/operator_anchors_aliases_test.go
@@ -320,6 +320,18 @@ somethingElse:
 
 var anchorOperatorScenarios = []expressionScenario{
 	{
+		// mergeObjects previously skipped all !!merge-tagged nodes. Since !!merge only appears on
+		// << map keys, this meant applyAssignment was never called for the << key. It was later
+		// autocreated by createStringScalarNode("<<") with tag !!str, silently dropping !!merge.
+		// DontFollowAlias:true already prevents aliases being followed, so the skip was redundant.
+		// Old (buggy) output: "D0, P[], (!!map)::base: &base\n    x: 1\ndest:\n    <<: *base\n"
+		skipDoc:     true,
+		description: "direct *+ preserves explicit !!merge tag on << key (regression for issue 2677)",
+		document:    "base: &base\n    x: 1\ndest:\n    !!merge <<: *base\n",
+		expression:  `. as $d | {} *+ $d`,
+		expected:    []string{"D0, P[], (!!map)::base: &base\n    x: 1\ndest:\n    !!merge <<: *base\n"},
+	},
+	{
 		skipDoc:     true,
 		description: "explicit !!merge tag on << key is preserved through ireduce merge",
 		document:    mixedMergeTagStyleDocument,
--- a/pkg/yqlib/operator_anchors_aliases_test.go
+++ b/pkg/yqlib/operator_anchors_aliases_test.go
@@ -288,7 +288,51 @@ var badAnchorOperatorScenarios = []expressionScenario{
 	},
 }
 
+var mixedMergeTagStyleDocument = `
+constants:
+    errorResponse: &errorResponse
+        status: 200
+endpoints:
+    - condition: true
+      !!merge <<: *errorResponse
+    - condition: false
+      <<: *errorResponse
+other:
+    !!merge <<: *errorResponse
+somethingElse:
+    <<: *errorResponse
+`
+
+var mixedMergeTagStyleExplodedDocument = `
+constants:
+    errorResponse:
+        status: 200
+endpoints:
+    - condition: true
+      status: 200
+    - condition: false
+      status: 200
+other:
+    status: 200
+somethingElse:
+    status: 200
+`
+
 var anchorOperatorScenarios = []expressionScenario{
+	{
+		skipDoc:     true,
+		description: "explicit !!merge tag on << key is preserved through ireduce merge",
+		document:    mixedMergeTagStyleDocument,
+		expression:  `. as $item ireduce ({}; . *+ $item)`,
+		expected:    []string{"D0, P[], (!!map)::" + mixedMergeTagStyleDocument},
+	},
+	{
+		skipDoc:     true,
+		description: "explode expands << merge keys regardless of explicit tag style (!!merge or plain)",
+		document:    mixedMergeTagStyleDocument,
+		expression:  `explode(.)`,
+		expected:    []string{"D0, P[], (!!map)::" + mixedMergeTagStyleExplodedDocument},
+	},
 	{
 		skipDoc:     true,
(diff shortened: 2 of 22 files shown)


pkg/yqlib/candidate_node.go
```diff
@@ -46,7 +46,11 @@ const (
 func createStringScalarNode(stringValue string) *CandidateNode {
 	var node = &CandidateNode{Kind: ScalarNode}
 	node.Value = stringValue
-	node.Tag = "!!str"
+	if stringValue == "<<" {
+		node.Tag = "!!merge"
+	} else {
+		node.Tag = "!!str"
+	}
 	return node
```
ccoVeille, line 54: Could you add tests for this?

Because it's not really obvious

W-Floyd: Added a test that replicates what's documented in #2677, will look at doing a more targeted test that exercises just this functionality.

W-Floyd:
Changed the existing tests to pass given the changes in https://github.com/yaml/go-yaml/pull/348, added a test for the given fix, and found a corner case to improve specific to the example with multiplication. I am still very new to both the `go-yaml` and `yq` codebases, and neither am I intimately familiar with the YAML spec, so sorry for the back and forth, and I appreciate the feedback!

***
It seems when `yq` autocreates a `<<` key it was creating it with `tag=!!str`, which `go-yaml` was handling for us before. Given the changes in http://github.com/yaml/go-yaml/commit/fbd745be58df058ec8eb7d7b42468458f8d8ec64, we need to give it `tag=!!merge` ourselves. Note that this is only for the nodes `yq` creates itself, hence why nested alias/anchor stuff (that were handled by `go-yaml` wholesale) kept working.

Regarding `mergeObjects` in `operator_multiply.go`, during a merge operation (`*+`), the code iterated over all nodes from the RHS and skipped any with `tag=!!merge`. It seems the intent was to avoid accidentally following merge anchors, but since `DontFollowAlias: true` is already set on the traversal, that protection was redundant. The problem was that skipping the `<<` key node meant yq never ran the key assignment for it — so instead of copying the source key's style, it autocreated a fresh one with `Style=0`, silently dropping the explicit `!!merge` annotation. The fix narrows the skip to `!candidate.IsMapKey`: values tagged `!!merge` (hypothetical) are still skipped, but `<<` key nodes are processed normally, and `UpdateFrom` copies both the tag and the style from the source.

ccoVeille:
@mikefarah @ingydotnet what do you think about what @W-Floyd explained ?

Thanks

I think his analysis is good, but it leads me to question the change we did in go-yaml with 

https://github.com/yaml/go-yaml/commit/fbd745be58df058ec8eb7d7b42468458f8d8ec64

It might be OK, and then imply a change in yq like the one suggested here, but I prefer everyone to review the changes here and the one we did in go-yaml

ccoVeille:
I'm very surprised by the commit

eb7738f828dac12bccff7f9cd0b989ff34d2e9c9

That remove the merge and float tags in a lot of files.

It doesn't sound OK.

Maybe it's due to the fact the [go-yaml PR needs](https://github.com/yaml/go-yaml/pull/348) to be merged first, but seeing such commit that changes so much things make me feel there is something wrong

W-Floyd:
Yes, it is for the sake of https://github.com/yaml/go-yaml/pull/348 - I can back out those changes if they should wait. But neither will the added test work until that's merged.

pkg/yqlib/operator_modulo_test.go
```diff
@@ -37,7 +37,7 @@ var moduloOperatorScenarios = []expressionScenario{
 		document:       `{a: 12, b: 2.5}`,
 		expression:     `.a = .a % .b`,
 		expected: []string{
-			"D0, P[], (!!map)::{a: !!float 2, b: 2.5}\n",
+			"D0, P[], (!!map)::{a: 2, b: 2.5}\n",
```
mikefarah, line 40: This seems to have change a whole bunch of unrelated things...

pkg/yqlib/operator_anchors_aliases_test.go
```diff
@@ -279,7 +279,51 @@ var badAnchorOperatorScenarios = []expressionScenario{
...
+		document:    mixedMergeTagStyleDocument,
+		expression:  `. as $item ireduce ({}; . *+ $item)`,
+		expected:    []string{"D0, P[], (!!map)::" + mixedMergeTagStyleDocument},
+	},
+	{
+		skipDoc:     true,
+		description: "explode expands << merge keys regardless of explicit tag style (!!merge or plain)",
+		document:    mixedMergeTagStyleDocument,
+		expression:  `explode(.)`,
+		expected:    []string{"D0, P[], (!!map)::" + mixedMergeTagStyleExplodedDocument},
```
mikefarah, line 346: This test makes sense to me 👍🏼

pkg/yqlib/operator_multiply.go
```diff
@@ -189,7 +189,7 @@ func mergeObjects(d *dataTreeNavigator, context Context, lhs *CandidateNode, rhs
 
 		log.Debugf("going to applied assignment to LHS: %v with RHS: %v", NodeToString(lhs), NodeToString(candidate))
 
-		if candidate.Tag == "!!merge" {
+		if candidate.Tag == "!!merge" && !candidate.IsMapKey {
```
mikefarah: been a long time since I've touched this logic...I'd be wary of making changes. Not sure why this is needed?

pkg/yqlib/operator_modulo_test.go
W-Floyd: Sorry, that's to match the new output of https://github.com/yaml/go-yaml/pull/348 - I can revert it for now if we want to handle that separately.

W-Floyd:
Testing by adding this to `go.mod`:
```
replace go.yaml.in/yaml/v4 v4.0.0-rc.4 => github.com/yaml/go-yaml/v4 v4.0.0-20260512181823-09a581dc5494
```

pkg/yqlib/operator_multiply.go
W-Floyd: I took more of a look, and believe this entire check is no longer required, due to [`DontFollowAlias: true`](https://github.com/W-Floyd/yq/blob/1044baa2958054bf4ffa2d2ef56001059c130c4f/pkg/yqlib/operator_multiply.go#L173). Removing it has the same result. [I added a test](https://github.com/mikefarah/yq/pull/2705/changes#diff-d9f4cbe5678e7f0ddc483038cd910cf82e5386730413de26314136e74d31b2dfR313-R324) to show what this accomplishes.

pkg/yqlib/operator_modulo_test.go
ccoVeille: This sounds strange to me then.

I have commented the go-yaml PR

https://github.com/yaml/go-yaml/pull/348#issuecomment-4457765287

pkg/yqlib/operator_multiply.go
```diff
@@ -189,10 +189,6 @@ func mergeObjects(d *dataTreeNavigator, context Context, lhs *CandidateNode, rhs
 
 		log.Debugf("going to applied assignment to LHS: %v with RHS: %v", NodeToString(lhs), NodeToString(candidate))
 
-		if candidate.Tag == "!!merge" {
```
mikefarah, line 192: why is this not needed?

mikefarah: Oh, got it,  read your comment later in the PR

ingydotnet:
While I believe that https://github.com/yaml/go-yaml/pull/348 is doing the "right" thing, It occurred to me later that the right thing doesn't necessarily mean the "same" thing as v3. I'll take a look and report back on that.

mikefarah:
Thanks @W-Floyd , @ccoVeille and @ingydotnet ! Will have this fix in the next yq release :)

mikefarah:
Fixed in 4.53.4

PR #2721, opened 2026-05-27
feat: add `in` operator (inverse of `has`)

### LINKED ISSUE #2322: Feature request: Membership operator (probably called `in`)
Inverse operator of `contains`, probably called `in` or something alike?

If we have data1.yml like:
(please keep to around 10 lines )

```yaml
- {item: 'Pizza', type: 'Food'}
- {item: 'Pasta', type: 'Food'}
- {item: 'Rose', type: 'Flower'}
- {item: 'Tulip', type: 'Flower'}
- {item: 'Hammer', type: 'Tool'}
- {item: 'Screwdriver', type: 'Tool'}
```

And we run a command:

```bash
yq 'filter(.type | in(["Tool", "Food"]))' data1.yml
```

it could output

```yaml
- {item: 'Pizza', type: 'Food'}
- {item: 'Pasta', type: 'Food'}
- {item: 'Hammer', type: 'Tool'}
- {item: 'Screwdriver', type: 'Tool'}
```

```bash
yq 'filter(.type as $el | ["Tool", "Food"] | contains([$el]))' data1.yml
```

**Additional context**
`jq` also has the [`in` operator](https://jqlang.org/manual/#in)

toller892:
## What

Add the `in` operator to yq, mirroring jq's `in` operator (inverse of `has`).

## Why

Closes #2322. The `in` operator checks whether a value is present in a mapping (as a key) or sequence (as an element), enabling idiomatic membership filtering in yq expressions.

## Usage

```yaml
# data.yml
- {item: Pizza, type: Food}
- {item: Rose, type: Flower}
- {item: Hammer, type: Tool}
```

```bash
# Filter by membership — select items whose type is Tool or Food
yq '.[] | select(.type | in(["Tool", "Food"]))' data.yml
```

```yaml
# Output
{item: Pizza, type: Food}
{item: Hammer, type: Tool}
```

```bash
# Check map key existence (using variable binding)
echo 'a: 1' | yq '. as $m | "a" | in($m)'
# true
```

## Implementation

- **`operator_in.go`** — Handler function: evaluates the RHS to get the collection, then checks if the LHS (piped-in value) is present as a key in a mapping or as an element in a sequence.
- **`operator_in_test.go`** — 11 test scenarios covering map key checks, array value membership, `select()` integration, and variable binding patterns.
- **`operation.go`** — Registers `inOpType` with `NumArgs: 1, Precedence: 50`.
- **`lexer_participle.go`** — Adds `simpleOp("in", inOpType)` to the lexer.

## Note on context model

Because RHS arguments are evaluated in the pipe context (not the root document), `in(.)` evaluates `.` as the piped value, not the original document. Use variable bindings to capture the collection first:

```bash
# Does NOT work as expected (`. evaluates to the pipe input):
echo 'a: 1' | yq '"a" | in(.)'  # false

# Works correctly (capture document in variable):
echo 'a: 1' | yq '. as $m | "a" | in($m)'  # true
```

This matches yq's existing context model (same as `has` with complex expressions).

### DIFF
--- a/pkg/yqlib/operator_in.go
+++ b/pkg/yqlib/operator_in.go
@@ -1,0 +1,59 @@
+package yqlib
+
+import (
+	"container/list"
+)
+
+func inOperator(d *dataTreeNavigator, context Context, expressionNode *ExpressionNode) (Context, error) {
+
+	log.Debugf("inOperation")
+	var results = list.New()
+
+	rhs, err := d.GetMatchingNodes(context.ReadOnlyClone(), expressionNode.RHS)
+
+	if err != nil {
+		return Context{}, err
+	}
+
+	var collection *CandidateNode
+	if rhs.MatchingNodes.Len() != 0 {
+		collection = rhs.MatchingNodes.Front().Value.(*CandidateNode)
+	} else {
+		// no collection provided, return false for all
+		for el := context.MatchingNodes.Front(); el != nil; el = el.Next() {
+			candidate := el.Value.(*CandidateNode)
+			results.PushBack(createBooleanCandidate(candidate, false))
+		}
+		return context.ChildContext(results), nil
+	}
+
+	for el := context.MatchingNodes.Front(); el != nil; el = el.Next() {
+		candidate := el.Value.(*CandidateNode)
+		isIn := false
+
+		switch collection.Kind {
+		case MappingNode:
+			// Check if candidate value exists as a key in the mapping
+			for index := 0; index < len(collection.Content); index = index + 2 {
+				key := collection.Content[index]
+				if key.Value == candidate.Value {
+					isIn = true
+					break
+				}
+			}
+		case SequenceNode:
+			// Check if candidate value is present in the sequence (value membership)
+			for _, element := range collection.Content {
+				if element.Value == candidate.Value {
+					isIn = true
+					break
+				}
+			}
+		default:
+			isIn = false
+		}
+
+		results.PushBack(createBooleanCandidate(candidate, isIn))
+	}
+	return context.ChildContext(results), nil
+}
--- a/pkg/yqlib/operator_in_test.go
+++ b/pkg/yqlib/operator_in_test.go
@@ -1,0 +1,105 @@
+package yqlib
+
+import (
+	"testing"
+)
+
+var inOperatorScenarios = []expressionScenario{
+	{
+		description: "Check key exists in map using variable binding",
+		document:    "a: 1\nb: 2\nc: 3\n",
+		expression:  `. as $m | "a" | in($m)`,
+		expected: []string{
+			"D0, P[], (!!bool)::true\n",
+		},
+	},
+	{
+		description: "Check key does not exist in map",
+		document:    "a: 1\nb: 2\nc: 3\n",
+		expression:  `. as $m | "d" | in($m)`,
+		expected: []string{
+			"D0, P[], (!!bool)::false\n",
+		},
+	},
+	{
+		description: "Check value exists in array",
+		document:    "- Tool\n- Food\n- Flower\n",
+		expression:  `. as $m | "Food" | in($m)`,
+		expected: []string{
+			"D0, P[], (!!bool)::true\n",
+		},
+	},
+	{
+		description: "Check value does not exist in array",
+		document:    "- Tool\n- Food\n- Flower\n",
+		expression:  `. as $m | "Animal" | in($m)`,
+		expected: []string{
+			"D0, P[], (!!bool)::false\n",
+		},
+	},
+	{
+		description:    "Check in with select on array elements",
+		subdescription: "Filter items whose type is in the given list",
+		document:       "- {item: Pizza, type: Food}\n- {item: Rose, type: Flower}\n- {item: Hammer, type: Tool}\n",
+		expression:     `.[] | select(.type | in(["Tool", "Food"]))`,
+		expected: []string{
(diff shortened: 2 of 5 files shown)


pkg/yqlib/operator_in.go
```diff
@@ -0,0 +1,59 @@
...
+				key := collection.Content[index]
+				if key.Value == candidate.Value {
+					isIn = true
+					break
+				}
+			}
+		case SequenceNode:
+			// Check if candidate value is present in the sequence (value membership)
+			for _, element := range collection.Content {
+				if element.Value == candidate.Value {
```
mikefarah, line 47: > Generated by Cursor, not the reviewer.

**Blocking:** Comparing only `.Value` breaks non-scalar membership. Mapping/sequence nodes have an empty `.Value`, so any object spuriously matches any array of objects:

```bash
yq -n '{"x": 99} | in([{"a": 1}, {"b": 2}])'  # true — should be false
```

Please use `contains(element, candidate)` here (same deep equality as the `contains` operator) instead of string `.Value` comparison.

pkg/yqlib/operator_in.go
```diff
@@ -0,0 +1,59 @@
...
+	for el := context.MatchingNodes.Front(); el != nil; el = el.Next() {
+		candidate := el.Value.(*CandidateNode)
+		isIn := false
+
+		switch collection.Kind {
+		case MappingNode:
+			// Check if candidate value exists as a key in the mapping
+			for index := 0; index < len(collection.Content); index = index + 2 {
+				key := collection.Content[index]
+				if key.Value == candidate.Value {
```
mikefarah, line 39: > Generated by Cursor, not the reviewer.

Minor: map key checks also compare only `.Value` (no tag check). This matches `has` behaviour for keys, which is fine — but worth noting that scalar array membership below should ideally align with `contains` tag semantics (`1` vs `"1"`).

pkg/yqlib/operator_in_test.go
```diff
@@ -0,0 +1,105 @@
...
+	{
+		description: "Check value does not exist in array",
+		document:    "- Tool\n- Food\n- Flower\n",
+		expression:  `. as $m | "Animal" | in($m)`,
+		expected: []string{
+			"D0, P[], (!!bool)::false\n",
+		},
+	},
+	{
+		description:    "Check in with select on array elements",
```
mikefarah, line 41: > Generated by Cursor, not the reviewer.

Please add test coverage for the non-scalar array bug, e.g.:

```go
{
    skipDoc:    true,
    expression: `{"x": 99} | in([{"a": 1}, {"b": 2}])`,
    expected:   []string{"D0, P[], (!!bool)::false\n"},
},
```

Also consider tests for tag mismatch (`1 | in(["1"])`) and `null` in arrays.

pkg/yqlib/operator_in_test.go
```diff
@@ -0,0 +1,105 @@
...
+		document:       "- {item: Pizza, type: Food}\n- {item: Rose, type: Flower}\n- {item: Hammer, type: Tool}\n",
+		expression:     `.[] | select(.type | in(["Tool", "Food"]))`,
+		expected: []string{
+			"D0, P[0], (!!map)::{item: Pizza, type: Food}\n",
+			"D0, P[2], (!!map)::{item: Hammer, type: Tool}\n",
+		},
+	},
+	{
+		description: "In with variable binding - found",
+		document:    "a: 1\nb: 2\nc: 3\n",
```
mikefarah, line 52: > Generated by Cursor, not the reviewer.

These "In with variable binding" scenarios duplicate "Check key exists/does not exist in map" above (lines 9–23). Consider removing the redundant pair to keep the test file lean.

mikefarah (changes requested):
> **Note:** This review was generated by [Cursor](https://cursor.com) (AI-assisted code review), not written manually by the reviewer.

## Summary

Thanks for this PR — the `in` operator is a good addition and addresses the primary use case in #2322 (`select(.type | in(["Tool", "Food"]))`). Registration, lexer wiring, and the test/doc pattern all follow existing yq conventions.

**Verdict: request changes** — there is a correctness bug in non-scalar array membership that should be fixed before merge.

## What's good

- Correct placement next to `has` (precedence 50, lexer registration)
- Primary filtering use case from #2322 works
- Honest documentation of the context model (`in(.)` vs `in($m)`)
- Tests pass (`TestInOperatorScenarios`, full `pkg/yqlib` suite)

## Blocking issue

**Non-scalar array membership is broken.** The sequence branch compares only `.Value`, so any mapping/sequence node (empty `.Value`) falsely matches any array containing objects. Example: `{"x": 99} | in([{"a": 1}])` returns `true` but should be `false`. `contains` already does deep comparison — the sequence branch should reuse it.

## Other items

1. **Missing doc header** — please add `pkg/yqlib/doc/operators/headers/in.md` (see `headers/has.md`, `headers/contains.md`).
2. **Scalar tag comparison** — `1 | in(["1"])` returns `true` because only `.Value` is compared; `contains` enforces tag matching. Worth aligning or documenting.
3. **Duplicate tests** — "Check key exists in map using variable binding" / "In with variable binding - found" (and the not-found pair) are redundant.
4. **Test gaps** — no coverage for non-scalar array membership, tag mismatch, or `null` in arrays.

## Suggested fix

```go
case SequenceNode:
    for _, element := range collection.Content {
        contained, err := contains(element, candidate)
        if err != nil {
            return Context{}, err
        }
        if contained {
            isIn = true
            break
        }
    }
```

Add a test asserting `{"x": 99} | in([{"a": 1}])` → `false`.

PR #2722, opened 2026-05-27
Fix download command for yq in README

addayounes:
Updated the download command for yq to use the correct URL format.


mikefarah:
That URL doesn't work

PR #2750, opened 2026-06-15
fix: apply per-candidate index when traversing arrays/maps by a streamed index

### LINKED ISSUE #2593: Extracting object values by iterating via keys causes missing items
**Describe the bug**
Extracting object values by iterating via keys causes missing items, see the example below:

Version of yq: 4.45.2
Operating system: mac
Installed via: homebrew

**Command**
The command you ran:
```
echo '["a","b"]' | yq '. as $o | keys[] | $o[.]'
```

**Actual behaviour**

```yaml
a
```

**Expected behaviour**

```yaml
a
b
```

**Additional context**
The problem appears to be related to the `$o[.]` part, somehow it only produced the value of the first key (index `0`).

However, it works when iterating over `keys[]` with an intermediate variable, e.g.:
- `echo '["a","b"]' | yq '. as $o | keys[] | . as $k | $o[$k]'`
- `echo '["a","b"]' | yq '. as $o | keys[] | . as $k | $o[.]'`   ## `$k` is not even used

Both would produce the expected result mentioned above.

StressTestor:
> **Note:** this PR was prepared by an AI agent (Claude Code) acting on behalf of @StressTestor, not by the user personally.

## what

`$o[.]` over a streamed context only returned the first match. this fixes it.

```sh
echo '["a","b"]' | yq '. as $o | keys[] | $o[.]'
# before: a
# after:  a
#         b
```

it hits maps the same way: `echo '{"x":1,"y":2}' | yq '. as $o | keys[] | $o[.]'` returned only `1`.

## why

the index in `$o[.]` is a collect expression that yields one index set per incoming candidate (for `keys[]` streaming `0` then `1`, it yields `[0]` then `[1]`). `traverseArrayOperator` evaluated it over the whole context but then used only `rhs.MatchingNodes.Front()`, the first index set, so every candidate but the first was dropped. the `. as $k | $o[.]` workaround worked because `as` splits the stream, so each call sees a single candidate and `Front()` is enough.

## fix

pair each index set with its candidate (`pkg/yqlib/operator_traverse_path.go`):

- when the LHS has one node per candidate (`.[] | .[idx]`), traverse each LHS node with its own index set
- when the LHS collapses to a single node (a variable like `$o`), traverse it against every index set

LHS and RHS are still each evaluated once over the whole context, so positional operators like `split_doc` keep their document index.

## tests

regression scenarios for the array and map forms in `operator_traverse_path_test.go`. full `go test ./...` passes, and the existing traverse and split-doc scenarios are unchanged.

Fixes #2593.

### DIFF
--- a/pkg/yqlib/operator_traverse_path.go
+++ b/pkg/yqlib/operator_traverse_path.go
@@ -111,10 +111,10 @@ func traverseArrayOperator(d *dataTreeNavigator, context Context, expressionNode
 		return Context{}, err
 	}
 
-	// rhs is a collect expression that will yield indices to retrieve of the arrays
-
+	// rhs is a collect expression that yields the indices to retrieve. It is
+	// evaluated over the whole context, producing one index set per incoming
+	// candidate.
 	rhs, err := d.GetMatchingNodes(context.ReadOnlyClone(), expressionNode.RHS)
-
 	if err != nil {
 		return Context{}, err
 	}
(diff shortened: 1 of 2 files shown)


ryenus:
This looks great, @mikefarah would you please take a look? Hopefully https://github.com/mikefarah/yq/issues/2593 can be fixed soon with this, thank you!

PR #2751, opened 2026-06-18
fix: skip UTF-8 BOM when processing front matter

### LINKED ISSUE #2496: Process front matter of UTF-8 files with BOM doesn't work.
**Describe the bug**
When processing front matter of UTF-8 with BOM files first `---` gets deleted.

Version of yq: 4.45.4
Operating system: linux
Installed via: binary release

**Input Yaml**

Create file with BOM
`printf '\xEF\xBB\xBF---\ntitle: Test\n---\n' > test_with_bom.md`

Value of file is:
```
cat test_with_bom.md
---
title: Test
---
```

yq deleted first line:
```
$ yq --front-matter=process test_with_bom.md
title: Test
---
```

**Command**
The command you ran:
```
yq --front-matter=process test_with_bom.md
```

**Actual behavior**

```yaml
title: Test
---
```

**Expected behavior**

```yaml
---
title: Test
---
```

StressTestor:
> Disclosure: this PR was written by an AI agent (Claude Code) acting on the user's behalf, not the user personally.

## what

`yq --front-matter=process` drops the opening `---` when the input file starts with a UTF-8 BOM, which loses data:

```
$ printf '\xEF\xBB\xBF---\ntitle: Test\n---\n' > test.md
$ yq --front-matter=process test.md
title: Test
---
```

The opening `---` is gone. This is #2496.

## why

`front_matter.go`'s `Split()` wraps the file in a `bufio.Reader` without skipping a leading BOM. The first line read is then `\xEF\xBB\xBF---`, so the `---` separator is not recognised and that line is treated as content instead of the front-matter delimiter.

## fix

Skip the BOM with `utfbom.Skip` before wrapping the reader, the same way `decoder_csv_object.go` already does. `utfbom` is already a direct dependency.

```go
cleanReader, _ := utfbom.Skip(file)
reader = bufio.NewReader(cleanReader)
```

Applied to both the file and stdin paths. `utfbom.Skip` replays non-BOM bytes unchanged, so files without a BOM are byte-identical.

## tests

`TestFrontMatterSplitWithBOM` in `front_matter_test.go` feeds a BOM-prefixed document and asserts the front matter and content split correctly. It fails before the fix (the front matter starts with the BOM) and passes after.

Fixes #2496.

### DIFF
--- a/pkg/yqlib/front_matter.go
+++ b/pkg/yqlib/front_matter.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"io"
 	"os"
+
+	"github.com/dimchansky/utfbom"
 )
 
 type frontMatterHandler interface {
@@ -43,13 +45,15 @@ func (f *frontMatterHandlerImpl) Split() error {
 	var reader *bufio.Reader
 	var err error
 	if f.originalFilename == "-" {
-		reader = bufio.NewReader(os.Stdin)
+		cleanReader, _ := utfbom.Skip(os.Stdin)
+		reader = bufio.NewReader(cleanReader)
 	} else {
 		file, err := os.Open(f.originalFilename) // #nosec
 		if err != nil {
 			return err
 		}
-		reader = bufio.NewReader(file)
+		cleanReader, _ := utfbom.Skip(file)
+		reader = bufio.NewReader(cleanReader)
 	}
 	f.contentReader = reader
 
--- a/pkg/yqlib/front_matter.go
+++ b/pkg/yqlib/front_matter.go
@@ -2,12 +2,25 @@ package yqlib
 
 import (
 	"bufio"
+	"bytes"
 	"errors"
 	"io"
 	"os"
-
-	"github.com/dimchansky/utfbom"
 )
+
+var utf8BOM = []byte{0xEF, 0xBB, 0xBF}
+
+// stripUTF8BOM returns a reader that skips a leading UTF-8 BOM, if present.
+func stripUTF8BOM(r io.Reader) io.Reader {
+	br := bufio.NewReader(r)
+
+	peek, err := br.Peek(3)
+	if err == nil && bytes.Equal(peek, utf8BOM) {
+		_, _ = br.Discard(3)
+	}
+
+	return br
+}
 
 type frontMatterHandler interface {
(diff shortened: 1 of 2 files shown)


mikefarah:
@cursoragent review this

pkg/yqlib/front_matter.go
```diff
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"io"
 	"os"
+
+	"github.com/dimchansky/utfbom"
```
ccoVeille: I find strange to add such dependency to skip UTF BOM.

The repository is stale (which is OK because of the feature stability), but as you use only Skip method.

Is there a need in yq to handle all UTF BOM. Isn't UTF-8 enough?

 
```
var utf8BOM = []byte{0xEF, 0xBB, 0xBF}

// StripUTF8BOM returns a reader that skips a leading UTF-8 BOM if present.
func StripUTF8BOM(r io.Reader) io.Reader {
	br := bufio.NewReader(r)

	peek, err := br.Peek(3)
	if err == nil && bytes.Equal(peek, utf8BOM) {
		_, _ = br.Discard(3)
	}

	return br
}
```

ccoVeille: What do you think @mikefarah

mikefarah: Yeah I agree ^

StressTestor:
reworked as suggested, thanks. front matter now uses a small unexported `stripUTF8BOM` helper (peek 3 bytes, discard if they're the UTF-8 BOM) on both the file and stdin paths, and the `utfbom` import is gone from that code. pushed as a separate commit so the delta is easy to review.

two notes:

- go.mod is untouched. `utfbom` was already a direct dependency before this PR and `decoder_csv_object.go` still uses it for CSV BOM detection, so this PR can't drop it outright. happy to do that removal as a follow-up if you want the dependency gone entirely.
- i kept the helper unexported to match the other internal helpers in yqlib (`readStream`, `safelyCloseFile`, etc). easy to export if you'd rather have it public.

also added a stdin-path regression test next to the existing file-path one, since both readers go through the helper now.

mikefarah:
Fixed in 4.53.4

PR #2771, opened 2026-07-06
fix(sort): avoid int64 overflow comparing large integers

maximilize:
> **Note:** this PR was prepared by an AI agent (Claude Code) acting on behalf of @maximilize, not by the user personally.

## what

`sort` orders large-magnitude integers incorrectly:

```sh
echo '[5000000000000000000, -5000000000000000000]' | yq 'sort' -o=json -I=0
# before: [5000000000000000000,-5000000000000000000]  (unsorted)
# after:  [-5000000000000000000,5000000000000000000]
```

## why

The int/int branch of the sort comparator (`pkg/yqlib/operator_sort.go`) returns `int(lhsNum - rhsNum)`. The subtraction is done in `int64` and overflows when the difference exceeds the int64 range (`5e18 - (-5e18) = 1e19` wraps negative), so the comparator reports the wrong sign.

## fix

Compare directly (`<` / `>`) instead of subtracting, mirroring the float branch just below. Regression scenario added (`skipDoc`).

### DIFF
--- a/pkg/yqlib/operator_sort.go
+++ b/pkg/yqlib/operator_sort.go
@@ -166,7 +166,12 @@ func (a sortableNodeArray) compare(lhs *CandidateNode, rhs *CandidateNode, dateT
 		if err != nil {
 			panic(err)
 		}
-		return int(lhsNum - rhsNum)
+		if lhsNum < rhsNum {
+			return -1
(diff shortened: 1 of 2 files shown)


mikefarah:
Fixed in 4.53.4

PR #2784, opened 2026-07-20
feat: add --stdin-filename flag to auto-detect the format of STDIN input

### LINKED ISSUE #1876: RFE: ability to specify input filename for stdin

In https://github.com/mikefarah/yq/issues/1484, the input/output mode choice was made defaulting dependent on the input filename. Thanks!

Now, if this would be also possible when processing stdin input, it'd be even nicer. For example, various editor formatter integrations work by piping input to external formatters through stdin. They may have a filename, but the file might not be saved anywhere yet, etc.

Other formatters tend to implement this with a cmdline option `--stdin <filename>` (example: [dprint](https://dprint.dev/cli/#formatting-standard-input)), or `--stdin-filename <filename>` (example: [ruff format](https://docs.astral.sh/ruff/formatter/)). I'd personally go with the latter, it's clearer.

Note that this is just an indication of the filename for format detection purposes. The file does not necessarily exist.

N/A.

**Additional context**

N/A.

hsdfat:
> **Disclosure:** this pull request — code, tests and description — was written by Claude Code, an AI agent acting on the user's behalf, not by the user personally. Raised in line with this repository's `AGENTS.md` disclosure requirement.

Fixes #1876.

## What

Adds a `--stdin-filename` flag. When input comes from STDIN, the given name is used for automatic input/output format detection, exactly as if the data had been read from that file:

```bash
$ echo '{ "mike": { "things": "cool" } }' | yq --stdin-filename thing.json
{
  "mike": {
    "things": "cool"
  }
}
```

The name is only ever passed to `filepath.Ext` — the file is never opened and **does not need to exist**, which is what makes this useful for editor and formatter integrations. This follows the naming suggested in the issue (`--stdin-filename`, as used by ruff) rather than dprint's `--stdin`.

## Why

Per the issue: an editor integration that wants yq to format a buffer currently has to hardcode yq's list of formats and pass the right `-p` itself. Handing yq the buffer's filename instead lets yq's own detection do the work, so integrations stay correct as yq gains formats.

## How

The production change is three lines in `configureInputFormat`:

```go
formatFilename := inputFilename
if inputFilename == "-" && stdinFilename != "" {
        formatFilename = stdinFilename
}
```

`formatFilename` feeds format detection; `inputFilename` is left untouched so the existing `inputFilename != "-"` backwards-compatibility check still behaves correctly. Everything else is flag registration, tests and the regenerated README flag table. Nothing in `pkg/yqlib` is touched.

Behaviour is unchanged when the flag is absent, an explicit `-p`/`--input-format` still wins, and the flag is ignored when a real file argument is supplied.

## Tests

- `cmd/utils_test.go` — 11 new cases in the existing `TestConfigureInputFormat`/`TestConfigureFormats` tables: stdin with and without the flag, known/unknown/absent extension, explicit-format-wins, and ignored-when-a-file-is-given.
- `acceptance_tests/inputs-format-auto.sh` — 5 e2e cases across `yq` and `yq ea`, covering JSON and XML stdin, unknown extensions, the explicit `-` argument form, and the `-p` + `--stdin-filename` combination.

`./scripts/format.sh`, `./scripts/check.sh`, `./scripts/spelling.sh`, `go test ./cmd/...` and `scripts/acceptance.sh` all pass.

Two failures exist on clean `master` and are unrelated to this change (confirmed by re-running there): `TestTomlScenarios` after the go-toml v2.4.2 bump in e2f1d5c changed an error string, and a `gci` lint hit in `pkg/yqlib/encoder_toml.go`.

## One question on scope

I kept this to format detection only, since that is what the issue asks for — it does not touch `yqlib.SetFilenameAlias()`, so the `filename` operator and `bad file '-'` errors still report `-` for STDIN.

Wiring it through would make those report the given name too, which is arguably the least surprising behaviour, but it goes beyond the request and widens the diff into `pkg/yqlib`. Happy to add it in this PR or a follow-up if you'd prefer that — just say which.

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,6 +433,7 @@ Flags:
       --shell-key-separator string      separator for shell variable key paths (default "_")
   -s, --split-exp string                print each result (or doc) into a file named (exp). [exp] argument must return a string. You can use $index in the expression as the result counter. The necessary directories will be created.
       --split-exp-file string           Use a file to specify the split-exp expression.
+      --stdin-filename string           filename to assume for input read from STDIN, used to automatically detect the format. The file does not need to exist.
       --string-interpolation            Toggles strings interpolation of \(exp) (default true)
       --tsv-auto-parse                  parse TSV YAML/JSON values (default true)
   -r, --unwrapScalar                    unwrap scalar, print the value with no quotes, colors or comments. Defaults to true for yaml (default true)
--- a/acceptance_tests/inputs-format-auto.sh
+++ b/acceptance_tests/inputs-format-auto.sh
@@ -31,6 +31,81 @@ EOM
   assertEquals "$expected" "$X"
 }
 
+testStdinFilenameJson() {
+  read -r -d '' expected << EOM
+{
+  "mike": {
+    "things": "cool"
+  }
+}
+EOM
+
+  X=$(echo '{ "mike" : { "things": "cool" } }' | ./yq --stdin-filename test.json)
+  assertEquals "$expected" "$X"
+
+  X=$(echo '{ "mike" : { "things": "cool" } }' | ./yq --stdin-filename test.json -)
+  assertEquals "$expected" "$X"
+
+  X=$(echo '{ "mike" : { "things": "cool" } }' | ./yq ea --stdin-filename test.json)
+  assertEquals "$expected" "$X"
+}
+
+testStdinFilenameUnknownExtension() {
+  read -r -d '' expected << EOM
+mike:
+  things: cool
+EOM
+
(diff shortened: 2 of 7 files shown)


hsdfat:
Closing this to keep the review queue tidy — it's been open nearly three weeks with no maintainer feedback and the CI approval gate was never triggered, so I'd rather not leave it lingering. The change is small and self-contained (adds a `--stdin-filename` flag so the format of piped STDIN can be auto-detected from a filename hint). Happy to reopen and  rebase if there's interest in the feature down the line. Thanks for yq!

PR #2787, opened 2026-07-27
Preserve integers outside int64 range when encoding JSON

hdimer:
### Bug

Converting an integer outside the `int64` range to JSON fails, even though the value is a valid integer in both YAML and JSON and round-trips fine as YAML:

```
$ echo '{a: 12345678901234567890}' | yq -o=json
Error: json: error calling MarshalJSON for type *yqlib.CandidateNode: ... strconv.ParseInt: parsing "12345678901234567890": value out of range

$ echo '{a: 12345678901234567890}' | yq -o=yaml
a: 12345678901234567890
```

`MarshalJSON` for a `!!int` scalar goes through `GetValueRep`, which parses into an `int64` and overflows.

### Fix

Emit `!!int` scalars verbatim when their text is already a valid JSON integer literal, mirroring the existing `jsonFloatLiteral` handling for floats. JSON permits arbitrary-precision integers, so `12345678901234567890` is preserved instead of erroring. Non-JSON forms (hex, a leading `+`, leading zeros) fall through to the existing path unchanged.

Added a test (`Encode json: integer outside int64 range`).

Scope note: very large *negative* integers (below `int64` min) are resolved to `!!float` by the underlying YAML decoder, so those are a separate matter and out of scope here.

Disclosure: written with AI assistance (Claude Code); I reviewed it and verified it against the repo's tests and golangci-lint.


pkg/yqlib/json_test.go
```diff
@@ -189,6 +189,14 @@ var jsonScenarios = []formatScenario{
...
 		scenarioType: "encode",
 	},
+	{
+		description:  "Encode json: integer outside int64 range",
+		skipDoc:      true,
+		input:        `{small: 42, big: 12345678901234567890}`,
+		indent:       0,
+		expected:     "{\"small\":42,\"big\":12345678901234567890}\n",
+		scenarioType: "encode",
+	},
```
mikefarah, line 231: This needs more tests to verify the conditions added, e.g "0", negative numbers, Is "-0" permitted in json, it looks like this would allow it 🤔

hdimer:
Thanks for the review @mikefarah. Rebased on master (kept your new hex/octal cases) and addressed the `-0` question: it's valid JSON but a non-canonical spelling of 0, so I now exclude it from the verbatim path and it falls back to `0`. Added encode tests for `0`, negative ints, `!!int -0` → `0`, and a negative value outside the int64 range.

One thing I hit while testing: a *bare* negative integer past int64 (e.g. `-12345678901234567890`) is resolved as `!!float` by the YAML parser, so it never reaches this `!!int` path and still loses precision, whereas positive ones resolve as `!!int` and are preserved. That asymmetry looks like a separate resolver issue - happy to open a follow-up if you'd like it fixed too.

PR #2793, opened 2026-08-01
fix: data race in InitExpressionParser

### LINKED ISSUE #2788: `yqlib` data race: `InitExpressionParser` initializes shared state on every `Evaluate` call
`yqlib` cannot currently be used from more than one goroutine: `go test -race` reports a data race on every run of a 20 line program that calls `NewStringEvaluator().Evaluate` concurrently. The `yq` CLI is unaffected, because `cmd/root.go` initializes the parser once at startup, so the fix looks like giving library callers that same guarantee with a `sync.Once`.

**Describe the bug**

Version of yq: 4.53.3 (also present in 4.52.4)
Operating system: mac (darwin/arm64), also reproduced on linux/arm64
Installed via: Go module, `github.com/mikefarah/yq/v4/pkg/yqlib`

`stringEvaluator.Evaluate` and `allAtOnceEvaluator.EvaluateNodes` both call `InitExpressionParser()` on every invocation, and that function assigns the package-level `ExpressionParser` behind a plain nil check:

```go
var ExpressionParser ExpressionParserInterface

func InitExpressionParser() {
	if ExpressionParser == nil {
		ExpressionParser = newExpressionParser()
	}
}
```

Two goroutines evaluating expressions can therefore run `newExpressionParser()` at the same time. `newParticipleLexer` compounds it by mutating the shared package-level `participleYqRules` slice in place (`yqRule.ParticipleTokenType = symbols[yqRule.Name]`), so a third goroutine can read `ParticipleTokenType` through `getYqDefinition` while it is being written.

These are pointer and struct field accesses rather than map accesses, so the Go runtime does not abort the process the way it does for a racy map. It corrupts quietly instead, which makes this easy to miss without `-race`.

**Reproducer**

There is no CLI command for this one, so in place of the template's yaml and command, here is a self-contained Go test. It fails on every run I have tried, 5 out of 5, reporting 4 or 10 race warnings depending on interleaving.

<details>
<summary>go.mod and main_test.go, plus one full run</summary>

`go.mod`:

```
module yqrace

go 1.26.4

require github.com/mikefarah/yq/v4 v4.53.3
```

`main_test.go`:

```go
package main

import (
	"sync"
	"testing"

	"github.com/mikefarah/yq/v4/pkg/yqlib"
)

func TestConcurrentStringEvaluator(t *testing.T) {
	pref := yqlib.YamlPreferences{Indent: 2}

	var wg sync.WaitGroup
	for range 32 {
		wg.Add(1)
		go func() {
			defer wg.Done()
			enc := yqlib.NewYamlEncoder(pref)
			dec := yqlib.NewYamlDecoder(pref)
			if _, err := yqlib.NewStringEvaluator().Evaluate(".a.b", "a:\n  b: value\n", enc, dec); err != nil {
				t.Error(err)
			}
		}()
	}
	wg.Wait()
}
```

Run with `go test -race -run TestConcurrentStringEvaluator .`:

```
==================
WARNING: DATA RACE
Write at 0x000103a64288 by goroutine 25:
  github.com/mikefarah/yq/v4/pkg/yqlib.newParticipleLexer()
      pkg/yqlib/lexer_participle.go:268
  github.com/mikefarah/yq/v4/pkg/yqlib.newExpressionParser()
      pkg/yqlib/expression_parser.go:25
  github.com/mikefarah/yq/v4/pkg/yqlib.InitExpressionParser()
      pkg/yqlib/lib.go:17
  github.com/mikefarah/yq/v4/pkg/yqlib.(*stringEvaluator).Evaluate()
      pkg/yqlib/string_evaluator.go:54

Previous write at 0x000103a64288 by goroutine 21:
  [same stack, different goroutine]
==================
WARNING: DATA RACE
Read at 0x000103a64288 by goroutine 12:
  github.com/mikefarah/yq/v4/pkg/yqlib.(*participleLexer).getYqDefinition()
      pkg/yqlib/lexer_participle.go:576
  github.com/mikefarah/yq/v4/pkg/yqlib.(*participleLexer).Tokenise()
      pkg/yqlib/lexer_participle.go:598
  github.com/mikefarah/yq/v4/pkg/yqlib.(*expressionParserImpl).ParseExpression()
      pkg/yqlib/expression_parser.go:30
==================
WARNING: DATA RACE
Write at 0x000103a47580 by goroutine 14:
  github.com/mikefarah/yq/v4/pkg/yqlib.InitExpressionParser()
      pkg/yqlib/lib.go:17
==================
Found 4 data race(s)
FAIL	yqrace	0.586s
```

</details>

**Actual behaviour**

`Found 4 data race(s)`, on 5 out of 5 runs. The two distinct addresses are the `ExpressionParser` variable itself and the shared rule structs behind `participleYqRules`.

**Expected behaviour**

No data race. The existing nil check suggests the intent is already one-time initialization, so I read this as the guard not being concurrency safe rather than concurrent use being out of scope. If concurrent use of `yqlib` is intended to be unsupported, that would be worth saying in the docs, since `InitExpressionParser` and `ExpressionParser` are both exported.

**Mechanism**

Every `Evaluate` performs the lazy init, so the window is not confined to a program's startup: it reopens on each call until one of them wins. A single `sync.Once` closes both sides at once, because the shared rule mutation happens inside the same constructor, and it also gives later readers a happens-before edge for `getYqDefinition`.

Relevant paths: `pkg/yqlib/lib.go` (`InitExpressionParser`, `ExpressionParser`), `pkg/yqlib/lexer_participle.go` (`newParticipleLexer`, `getYqDefinition`, `participleYqRules`), `pkg/yqlib/string_evaluator.go`, `pkg/yqlib/all_at_once_evaluator.go`, `cmd/root.go`.

**Additional context**

The CLI does not hit this: `cmd/root.go` calls `InitExpressionParser()` once during setup, before any evaluation, so only library consumers are exposed.

#2282 is an open question about proper library usage, and the code there calls `InitExpressionParser()` explicitly before use. A `sync.Once` would make that call a convenience rather than a requirement, which seems closer to what a caller would expect from an exported initializer.

I found this through a downstream consumer, cloudposse/atmos, which evaluates expressions from one goroutine per config file.

Could `InitExpressionParser` wrap the assignment in a `sync.Once`? Happy to raise a PR with a test, as CONTRIBUTING asks.

ankit090701:
## What's the bug

`yqlib` cannot currently be used from more than one goroutine: `go test -race` reports a data race on every run of a small program that calls `NewStringEvaluator().Evaluate` concurrently. The `yq` CLI itself is unaffected, because `cmd/root.go` calls `InitExpressionParser()` once at startup, before any evaluation - only library consumers hit this.

## Root cause

`InitExpressionParser` (`pkg/yqlib/lib.go`) used an unsynchronized nil-check "lazy init" pattern:

```go
func InitExpressionParser() {
	if ExpressionParser == nil {
		ExpressionParser = newExpressionParser()
	}
}
```

Two goroutines calling it concurrently can both observe `ExpressionParser` as `nil` and both call `newExpressionParser()`, racing on the write to the package-level `ExpressionParser` variable.

`newExpressionParser()` also calls `newParticipleLexer()`, which mutates the shared package-level `participleYqRules` slice's elements in place:

```go
for _, yqRule := range participleYqRules {
	yqRule.ParticipleTokenType = symbols[yqRule.Name]
}
```

so a third goroutine reading those same fields via `getYqDefinition` can race too. These are plain pointer/struct field accesses rather than map accesses, so the race corrupts state quietly instead of the process aborting the way it would for a racy map - easy to miss without `-race`.

## The fix

Wrap the assignment in a `sync.Once` so it happens exactly once, with a proper happens-before edge for every later reader:

```go
var initExpressionParserOnce sync.Once

func InitExpressionParser() {
	initExpressionParserOnce.Do(func() {
		ExpressionParser = newExpressionParser()
	})
}
```

`ExpressionParser` stays a public package-level variable for backward compatibility, since it's read directly by several callers (`operator_eval.go`, `operator_expression.go`, `operator_strings.go`, `stream_evaluator.go`, `cmd/utils.go`), not just through `InitExpressionParser()`.

## Testing

- Added `TestInitExpressionParser_ConcurrentEvaluate` (`pkg/yqlib/lib_test.go`), which spawns 32 goroutines each calling `NewStringEvaluator().Evaluate` concurrently - matching the reporter's own reproducer almost exactly.
- Verified with `go test -race`: **without** the fix this reliably reports both races described above (confirmed by temporarily reverting just the `lib.go` change and rerunning); **with** the fix it passes cleanly across repeated runs (`-count=5`, and again in isolation).
- Ran the full `go test -race ./pkg/yqlib/... ./cmd/...` suite. The only two failures (`TestLoadScenarios`, `TestTomlScenarios`) are unrelated to this change and pre-existing - confirmed by reproducing them identically on an unmodified checkout: one is a `\r\n` vs `\n` line-ending difference from my Windows checkout, and the other is a wording difference in the vendored TOML parser's error message (`"basic string not terminated by \""` vs `"unterminated basic string"`), from dependency drift, not from anything this PR touches. (I don't have a Go toolchain on this machine, so everything was built/tested via Docker.)

Fixes #2788

### DIFF
--- a/pkg/yqlib/lib.go
+++ b/pkg/yqlib/lib.go
@@ -8,14 +8,21 @@ import (
 	"math"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 var ExpressionParser ExpressionParserInterface
 
+var initExpressionParserOnce sync.Once
+
+// InitExpressionParser initialises the package-level ExpressionParser, if it
+// hasn't been already. Safe to call from multiple goroutines concurrently:
+// the parser is built at most once, and every caller observes the same
+// fully-initialised instance.
 func InitExpressionParser() {
-	if ExpressionParser == nil {
+	initExpressionParserOnce.Do(func() {
 		ExpressionParser = newExpressionParser()
-	}
+	})
 }
 
 var log = newLogger()
--- a/pkg/yqlib/lib_test.go
+++ b/pkg/yqlib/lib_test.go
@@ -3,6 +3,7 @@ package yqlib
 import (
 	"fmt"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/mikefarah/yq/v4/test"
@@ -13,6 +14,27 @@ func TestGetLogger(t *testing.T) {
 	if l != log {
 		t.Fatal("GetLogger should return the yq logger instance, not a copy")
 	}
+}
+
+// Regression test for https://github.com/mikefarah/yq/issues/2788: calling
+// InitExpressionParser (directly, or indirectly via a StringEvaluator) from
+// multiple goroutines concurrently used to race on the package-level
+// ExpressionParser variable and on shared lexer rule state mutated while
+// building it. Run with -race to check.
+func TestInitExpressionParser_ConcurrentEvaluate(t *testing.T) {
+	var wg sync.WaitGroup
+	for range 32 {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			encoder := NewYamlEncoder(ConfiguredYamlPreferences)
+			decoder := NewYamlDecoder(ConfiguredYamlPreferences)
+			if _, err := NewStringEvaluator().Evaluate(".a.b", "a:\n  b: value\n", encoder, decoder); err != nil {
+				t.Error(err)
+			}
+		}()
+	}
+	wg.Wait()
 }
 
 type parseSnippetScenario struct {


mikefarah:
Thanks for raising, this is a duplicate of https://github.com/mikefarah/yq/pull/2789

PR #2813, opened 2026-08-14
fix: map errors on a scalar input

### LINKED ISSUE #2368: `map` operator on scalars should error
**Describe the bug**
Running the map operator on scalars should error out

Taken from https://github.com/mikefarah/yq/issues/2359

Version of yq: 4.45.4

```bash
yq -n '"" | map(.)'
```
 yields empty, but maybe it should error like jq

VXNCXNX:
> Generated by an AI agent (Claude Code) acting on the user's behalf, not the user personally.

Fixes #2368.

## The problem

`map` accepted scalars silently, and the output was misleading in two different ways depending on the type:

```
$ yq -n '3 | map(.)'
[]
$ yq -n '"cat" | map(.)'
cat
$ yq -n 'null | map(.)'
[]
```

The string case is the worst of the three: `map` isn't even returning a list, it's echoing the input back. So a typo in a path expression, where the path lands on a scalar rather than the collection you meant, produces plausible-looking output instead of telling you the path was wrong.

jq errors on all three:

```
$ jq -n 'null|map(.)'   → Cannot iterate over null (null)
$ jq -n '"cat"|map(.)'  → Cannot iterate over string ("cat")
$ jq -n '3|map(.)'      → Cannot iterate over number (3)
```

## The fix

Reject scalar input in `mapOperator`:

```
$ yq -n '"" | map(.)'
Error: cannot map !!str, map only works for maps and arrays
```

The wording follows `operator_keys.go`, which already phrases this class of rejection as *"cannot get keys of %v, keys only works for maps and arrays"*.

Note this also makes `null | map(.)` an error, since null is a scalar node. That matches jq, and it's the same misleading-success case: `.nonexistent | map(...)` currently returns `[]` as though the map ran.

Collections are unaffected: `[1,2] | map(.+1)`, `{a: 1} | map(.+1)` and `[] | map(.)` all behave exactly as before.

## Tests

Three scenarios added to `mapOperatorScenarios`, covering a string document, a piped empty string (the exact repro from the issue) and an int. All are `skipDoc: true`, matching how other operators handle their error scenarios (e.g. `operator_divide_test.go`), so `doc/operators/map.md` is byte-identical.

```
go build ./...                                    exit 0
go test ./pkg/yqlib/... -run TestMapOperatorScenarios   ok
```

Full `go test ./pkg/yqlib/...` has one failure, `TestTomlScenarios`, both before and after this change, a go-toml error-string mismatch (`basic string not terminated by "` vs `unterminated basic string`) unrelated to this PR.

## One thing I left alone

`map_values` has the same gap (`yq -n '"" | map_values(.)'` still succeeds silently with empty output), and jq errors there too. It's a separate function with a different code path, and your issue is specifically about `map`, so I didn't expand the scope. Happy to send a follow-up if you want it fixed the same way.


VXNCXNX:
Rebased both onto current `master`, 25 commits of drift each, no conflicts. `go test ./...` is clean on the rebased base for both.

One caveat: the push moved CI into `action_required`, so there is no signal on the PR page until someone approves the run. That is worse to look at than the stale result it replaced, though the stale one was measuring a base from four weeks ago.

- #2812, `+12/-0`: collecting documents under `eval-all` duplicates every result. `printf '{"a":1}\n{"a":2}\n' | yq -p=json -o=json ea '[.] | map(. * {"b":0})'` gives each entry twice, and three documents give nine. `collectTogether` reparents decoded nodes into a sequence but leaves `EvaluateTogether` set, so `map` takes the evaluate-all-together branch and re-evaluates the RHS once per candidate. An array element is not a stream document, so the flag is cleared when the result becomes a child.
- #2813, `+22/-0`, fixes #2368: `map` accepts scalars silently and misleads differently per type, `3 | map(.)` gives `[]` while `"cat" | map(.)` gives `cat`.

Both are small and independent.

PR #2821, opened 2026-08-18
Don't clobber existing values when merging with null

### LINKED ISSUE #2224: merge empty array
**Describe the bug**
unexpected merge array if we have empty array

Version of yq: 4.44.6
Operating system: linux

**Input Yaml**
data1.yml:
```yaml
list:
  - 1
```

data2.yml:
```yaml
list:
```

data3.yml:
```yaml
list:
  - 3
```

**Command**
The command you ran:
```
yq ea '. as $item ireduce ({}; . *+ $item )' *.yml > result.yml
```

**Actual behavior**

result.yml
```yaml
list:
  - 3
```

**Expected behavior**

```yaml
list:
  - 1
  - 3
```

**Additional context**
entry `-1` has been removed

but if we exclude `data2.yml` file (with `yq ea '. as $item ireduce ({}; . *+ $item )' data1.yml  data3.yml > result.yml`), that works as expected.

yhuikzdtguioaert:
**Describe the bug**

Merging a value against `null` overwrites the existing value instead of leaving it in place:

```bash
$ yq -n '{"a": 1} * {"a": null}'
a: null      # expected: a: 1
```

This is inconsistent with merging against a top-level `null`, which already keeps the left-hand side:

```bash
$ yq -n '{"a": 1} * null'
a: 1
```

It's especially surprising when reducing several documents together and one of them leaves a key empty (#2224):

`data1.yml` -> `list: [1]`, `data2.yml` -> `list:`, `data3.yml` -> `list: [3]`

```bash
$ yq ea '. as $item ireduce ({}; . *+ $item )' data1.yml data2.yml data3.yml
list:
  - 3          # the "1" from data1 has been dropped
```

**Fix**

When the right-hand side of a merge assignment is `null`, only write it when the target itself is `null`. Existing values are left untouched, while brand new keys are still added as `null`. This reuses the existing `OnlyWriteNull` assignment preference (the same mechanism behind the `n` merge flag), so the change is small and localised to `applyAssignment`.

After the fix:

```bash
$ yq -n '{"a": 1} * {"a": null}'
a: 1

$ yq ea '. as $item ireduce ({}; . *+ $item )' data1.yml data2.yml data3.yml
list:
  - 1
  - 3
```

Non-null values still overwrite as before, and merging brand new (null) keys still adds them.

**Tests / docs**

Added a documented scenario plus regression tests in `pkg/yqlib/operator_multiply_test.go`, and regenerated `pkg/yqlib/doc/operators/multiply-merge.md` via `go test -run TestMultiplyOperatorScenarios`.

Fixes #2224

### DIFF
--- a/pkg/yqlib/doc/operators/multiply-merge.md
+++ b/pkg/yqlib/doc/operators/multiply-merge.md
@@ -598,3 +598,22 @@ will output
 - some
 ```
 
+## Merge does not overwrite existing values with null
+A `null` value on the right hand side does not clobber the value on the left, matching the behaviour of merging with an explicit `null`. New keys are still added.
+
+Given a sample.yml file of:
+```yaml
+a: cat
+b: dog
+```
+then
+```bash
+yq '. * {"a": null, "c": null}' sample.yml
+```
+will output
+```yaml
+a: cat
+b: dog
+c: null
+```
+
--- a/pkg/yqlib/operator_multiply.go
+++ b/pkg/yqlib/operator_multiply.go
@@ -205,7 +205,16 @@ func applyAssignment(d *dataTreeNavigator, context Context, pathIndexToStartFrom
 	lhsPath := rhs.GetPath()[pathIndexToStartFrom:]
 	log.Debugf("merge - lhsPath %v", lhsPath)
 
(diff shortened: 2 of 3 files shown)


mikefarah:
Disagree, if you specify null it should override.
It's what jq does as well
```
jq -n '{"a": 1} * {"a": null}'
{
  "a": null
}
```

PR #2849, opened 2026-09-01
Preserve file permissions on in-place edits

### LINKED ISSUE #2845: Command 'yq -i '.setting = "value"' config.yml' deletes permissions for yaml files on Windows 11
**Describe the bug**
When we run command 'yq -i '.setting = "value"' config.yml' under other user (not admin) on Windows 11, 2019, 2022 config.yml file loses permission for this user. And we get 'Access denied' error

Version of yq: 4.53.6
Operating system: Windows Server 2019, 2022, Windows 11 23H2, 25H2
Installed via: chocolatey

**Command**
'yq -i '.setting = "value"' config.yml'

**Actual behaviour**
setting has not changed
permissions for config.yml file for user are deleted. And we get 'Access denied' error
config.yml
setting: old-value
```

**Expected behaviour**
permissions for config.yml file for user stay the same
config.yml
setting: value
```

**Additional context**
Previous versions didn't have such bug on Windows 2019, 2022, 11-23H2, 11-25H2

Jorge-Polanco-Roque:
Fixes #2845.

## Problem
`yq -i file.yaml` (in-place edit) does not preserve the target file's permissions / ACLs.

## Root cause
The temp file for in-place writes was created in the **system temp dir**, then renamed over the target. The rename brings the temp file's permissions/ACLs with it, dropping the original file's — notably on Windows, where ACLs are inherited from the parent directory rather than copied by `Chmod`.

## Approach
Create the temp file in the **target file's directory** so it inherits that directory's permissions/ACLs; the rename then preserves them. `createTempFile` gains a `dir` parameter (`""` keeps the previous system-temp behavior, used by the front-matter path). The in-place handler passes `filepath.Dir(inputFilename)`.

## Testing
- Added `TestWriteInPlaceHandlerImpl_CreateTempFile_SameDirAsTarget` — asserts the temp file lands in the target's directory (fails without the fix).
- `go build ./...` · `go vet ./format/` · `go test ./format/` green.

Note: ACL inheritance is Windows-specific; this is verified by the temp-dir mechanism (cross-platform) rather than by asserting real ACLs.

### DIFF
--- a/pkg/yqlib/file_utils.go
+++ b/pkg/yqlib/file_utils.go
@@ -72,10 +72,19 @@ func safelyCloseFile(file *os.File) {
 	}
 }
 
-func createTempFile() (*os.File, error) {
-	_, err := os.Stat(os.TempDir())
+// createTempFile creates a temp file in dir. When dir is "" the system temp
+// directory is used. For in-place edits, pass the target file's directory so
+// the temp file inherits that directory's permissions/ACLs (on Windows ACLs are
+// inherited from the parent dir, not copied by Chmod), preserving them across
+// the rename that replaces the original file.
+func createTempFile(dir string) (*os.File, error) {
+	tempDir := dir
+	if tempDir == "" {
+		tempDir = os.TempDir()
+	}
+	_, err := os.Stat(tempDir)
 	if os.IsNotExist(err) {
-		err = os.Mkdir(os.TempDir(), 0700)
+		err = os.Mkdir(tempDir, 0700)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


Jorge-Polanco-Roque:
Gentle ping — open about a week now. This preserves the original file permissions on in-place edits (`-i`) instead of resetting them. Glad to tweak anything on request. Thanks for maintaining yq!

Jorge-Polanco-Roque:
Going to close this one to keep my open PRs tidy. The change still stands (preserving file permissions on in-place `-i` edits instead of resetting them) if it's ever useful down the line — happy to reopen or resubmit. Thanks for yq!

PR #2850, opened 2026-09-01
fix: preserve explicit assignments in read-only expressions

### LINKED ISSUE #2848: *? merge operator seems to be broken in 4.53.6 if left-hand side is constructed
**Describe the bug**

The *? operator ("only merge existing fields") produces an empty object if the object on the left-hand side of the merge was constructed using the assignment operator. This used to work correctly through at least 4.44.1.

```shell
$ yq -n '({} | .a.b=3)'
a:
  b: 3

$ yq '({} | .a.b=3) *? .' - <<EOF
a:
  b: 1
  c: 2
EOF
{}
$ #^- expected .a.b to be set to 1

$ yq --version
yq (https://github.com/mikefarah/yq/) version v4.53.6
```
Version of yq: 4.53.6
Operating system: linux
Installed via: binary release

RRXXZZYY:
> AI agent disclosure: This pull request was prepared and submitted by Codex acting on RRXXZZYY's behalf, not by the account owner personally.

## Summary

Fixes #2848.

Binary operators evaluate their operands with automatic path creation disabled. That read-only setting also reached an explicit assignment inside an operand, so `({} | .a.b = 3)` stayed empty when used on the left side of `*?`.

Evaluate assignment targets with a writable context clone. Ordinary operand traversal remains read-only, while an explicit assignment can create its target as it does outside a binary expression.

## Test Plan

- `go test ./pkg/yqlib -run '^(TestMultiplyOperatorScenarios|TestAssignOperatorScenarios|TestPipeOperatorScenarios)$' -count=1`
- `go test . ./cmd ./test -count=1`
- `go test ./pkg/yqlib -skip '^(TestLoadScenarios|TestSystemOperatorEnabledScenarios|TestWriteInPlaceHandlerImpl_FinishWriteInPlace_Symlink_Success)$' -count=1`
- `scripts/format.sh`
- `scripts/spelling.sh`
- `scripts/check.sh` (0 issues)
- Built the CLI and verified the issue reproducer now returns only `a.b: 1`; the existing read-only regression still leaves `x: null`.

The unfiltered suite was also attempted on Windows. The three skipped tests fail unchanged on the upstream commit because of CRLF expectations, a Git Bash executable path containing spaces, and unavailable symlink privileges.

### DIFF
--- a/pkg/yqlib/operator_assign.go
+++ b/pkg/yqlib/operator_assign.go
@@ -30,7 +30,7 @@ func getAssignPreferences(preferences interface{}) assignPreferences {
 }
 
 func assignUpdateOperator(d *dataTreeNavigator, context Context, expressionNode *ExpressionNode) (Context, error) {
-	lhs, err := d.GetMatchingNodes(context, expressionNode.LHS)
+	lhs, err := d.GetMatchingNodes(context.WritableClone(), expressionNode.LHS)
 	if err != nil {
 		return Context{}, err
 	}
--- a/pkg/yqlib/operator_multiply_test.go
+++ b/pkg/yqlib/operator_multiply_test.go
@@ -440,6 +440,15 @@ var multiplyOperatorScenarios = []expressionScenario{
 		},
 	},
 	{
+		description: "Merge only existing fields into a constructed object",
+		skipDoc:     true,
+		document:    `{a: {b: 1, c: 2}}`,
+		expression:  `({} | .a.b = 3) *? .`,
+		expected: []string{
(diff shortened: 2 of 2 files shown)


mikefarah:
Makes sense, this fix also needs to apply to assignAttributesOperator in the same file

RRXXZZYY:
Applied the same writable-LHS context to `assignAttributesOperator` as suggested. I added a focused regression scenario for nested style assignment on a constructed object and synced the branch with current `master`.

Focused verification passes:
- `TestStyleOperatorScenarios`
- `TestMultiplyOperatorScenarios`

No unrelated files were changed.

PR #2857, opened 2026-09-06
Quote unwrapped top-level strings that would not round-trip

### LINKED ISSUE #2608: A single string scalar as output is not quoted if necessary (breaking roundtrip safety)

w1977-0:
Fixes #2608.

## Problem

With the default `--unwrapScalar` behavior, a top-level string scalar is written verbatim with no quoting — even when the plain form would re-parse as a completely different YAML value:

```console
$ echo '"this: should really work"' | yq eval '.' -
this: should really work      # re-parses as a map!
$ echo '"123"' | yq eval '.' -
123                          # re-parses as an int!
```

Piping the output back into yq (or any YAML consumer) silently changes the type, breaking round-trip safety. Even an explicitly double-quoted input string loses its quotes on the way out.

## Solution

Before taking the plain write path for a top-level scalar, re-parse its value and only write it verbatim when the result is a string scalar with the same value. Otherwise, fall through to the regular encoder path, which already emits the appropriate quoting. This is the re-parse check suggested in the issue discussion.

Intentionally kept unchanged:
- plain strings (`hello world`) — still written bare
- empty strings — still written as an empty line (too much legacy tooling depends on it to change here)
- values containing control characters (e.g. NUL) — keep going through the existing NUL-separated output check

## Behavior changes (visible, deliberate)

Strings whose bare form would re-parse as something else are now quoted. This surfaces in a few existing outputs, most notably `tag` names:

```console
$ yq eval 'tag' sample.yml
'!!map'        # was: !!map — which re-parses as a tagged node, not a string
```

Affected test expectations (tag/boolean/value operators, XML/base64-derived scalars, goccy) and the auto-generated operator docs are updated accordingly.

## Tests

- new regression cases in `yaml_test.go` covering the issue scenarios: map-like/int-like/bool-like strings quoted, plain strings bare, empty string bare
- updated 15 expectations across tag/boolean/value operators, XML, goccy
- `go test ./...`: all packages pass


mikefarah:
Breaks backwards compatability

PR #2869, opened 2026-09-17
fix: apply ireduce per pipeline input like jq reduce

### LINKED ISSUE #2859: possible bug with ireduce compared to jq reduce
Quoting:
< https://mikefarah.gitbook.io/yq/operators/reduce#yq-vs-jq-syntax >

> Reduce syntax in yq is a little different from jq - as yq (currently) isn't as sophisticated as jq and its only supports infix notation (e.g. a + b, where the operator is in the middle of the two parameters) - where as jq uses a mix of infix notation with prefix notation (e.g. reduce a b is like writing + a b).

First, there is a minor typo above.  I suggest removing the "s" in "its":
> ... as jq and it(s) only supports ...

When you concede in this quote that `jq` is more sophisticated, I'm unsure if you meant by `reduce` syntax only or also by implementation when compared to `ireduce`, so I _may_ have found a bug here.

I'm running latest version of `yq` on Linux/x86_64:

```
yq --version
yq (https://github.com/mikefarah/yq/) version v4.53.6
```

The first command from each shell (`bash`) pipeline below is the same.

This one produces the expected output:

```
yq -n -ojson '"\"a\"=\"b\" \"x\"=\"y\""' | yq -pjson -ojson '[capture("\"(?<key>.*?)\"=\"(?<value>.*?)\""; "g")]'
[
  {
    "key": "a",
    "value": "b"
  },
  {
    "key": "x",
    "value": "y"
  }
]
```
								      
This is consistent with `jq`; i.e. output is the same if `yq -pjson -ojson` from second command of shell pipeline is replaced with `jq`.

There is a difference between `jq` and `yq` when `capture` is expressed in terms of `match` and `(i)reduce` to construct a logically equivalent expression:

```
yq -n -ojson '"\"a\"=\"b\" \"x\"=\"y\""' | jq '[match("\"(?<key>.*?)\"=\"(?<value>.*?)\""; "g") | (reduce .captures[] as $capture ({}; .[$capture.name] = $capture.string))]'						      
[
  {
    "key": "a",
    "value": "b"
  },
  {
    "key": "x",
    "value": "y"
  }
]
```

```						     						      
yq -n -ojson '"\"a\"=\"b\" \"x\"=\"y\""' | yq -pjson -ojson '[match("\"(?<key>.*?)\"=\"(?<value>.*?)\""; "g") | (.captures[] as $capture ireduce ({}; .[$capture.name] = $capture.string))]'
[
  {
    "key": "x",
    "value": "y"
  }
]
```

`jq` produces expected output but `yq` does not.

00200200:
Fixes #2859

Problem: `capture("..."; "g")` already returns one object per match, but `match | ireduce .captures[]` folded both matches into one accumulator, so only the last pair survived.

Change: Run `ireduce` once per pipeline input, matching jq's `reduce` and the existing `as`/`collect` split. File-loaded `eval-all` merges stay combined.

Test: Added a case that expects both capture objects; existing `eval-all` merge tests still pass.


00200200:
This handles the pipeline input evaluation correctly and brings the behavior in line with jq. LGTM! 👍

PR #2873, opened 2026-09-20
fix: bump go.yaml.in/yaml/v4 to fix blank line inserted in folded scalars

### LINKED ISSUE #2409: Unexpected blank lines inserted into YAML result
**Describe the bug**
When evaluating a query, phantom blank lines are inserted into the result.

Note that any how to questions should be posted in the discussion board and not raised as an issue.

Version of yq: 4.45.4
Operating system: Linux
Installed via: releases binary

**Input Yaml**
meh.yml:
```yaml
env:
  YQ_QUERY: >-
    (.runs.steps // .jobs.*.steps) | map(
      select(.uses | test("@(v\d+(\.\d+)*|\d+(\.\d+)+|[a-f0-9]{40})$") | not)
      | .uses
      | "\(key | line) \(split(\"@\")[1])"
    )[]
```

**Command**
The command you ran:
```
yq .env.YQ_QUERY meh.yaml
```

**Actual behavior**

A blank line is inserted in the middle of and at the end variable value:

```shell
YQ_QUERY: >-
  (.runs.steps // .jobs.*.steps) | map(

    select(.uses | test("@(v\d+(\.\d+)*|\d+(\.\d+)+|[a-f0-9]{40})$") | not)
    | .uses
    | "\(key | line) \(split(\"@\")[1])"
  )[]

```

**Expected behavior**

```shell
YQ_QUERY: >-
  (.runs.steps // .jobs.*.steps) | map(
    select(.uses | test("@(v\d+(\.\d+)*|\d+(\.\d+)+|[a-f0-9]{40})$") | not)
    | .uses
    | "\(key | line) \(split(\"@\")[1])"
  )[]
```

**Additional context**

I thought maybe there was some weird/hidden character in the output, so I removed all the characters between `map(` and `select(`. This is the resulting YAML (`meh2.yaml):

```yaml
env:
  YQ_QUERY: >-
    (.runs.steps // .jobs.*.steps) | map(select(.uses | test("@(v\d+(\.\d+)*|\d+(\.\d+)+|[a-f0-9]{40})$") | not)
      | .uses
      | "\(key | line) \(split(\"@\")[1])"
    )[]
```

This is what `yq .env meh2.yaml` returns now:

```shell
YQ_QUERY: >-
  (.runs.steps // .jobs.*.steps) | map(select(.uses | test("@(v\d+(\.\d+)*|\d+(\.\d+)+|[a-f0-9]{40})$") | not)

    | .uses
    | "\(key | line) \(split(\"@\")[1])"
  )[]

```

Notice the blank line in the middle of the result and at the end. So it seems unrelated to the data since a location that didn't have blank lines before spontaneously showed them.

### LINKED ISSUE #2452: Sorting keys with ICU Message adds empty line
**Describe the bug**

I read the docs of Symfony about plural translations https://symfony.com/doc/current/reference/formats/message_format.html and created file with my message. Also created script for my CI/CD pipeline to run `yq sort_lines(..) $file` on every file in translations directory. I spotted an issue about adding random blank line on every app run so pipeline always fails

Version of yq: 4.47.1
Operating system: mac
Installed via: homebrew

**Input Yaml**

messages+intl-icu.en.yaml:
```yaml
common:
  age_year: >-
    {count, plural,
      one {# year}
      other {# years}
    }
```

**Command**
The command you ran:
```
yq eval 'sort_keys(..)'messages+intl-icu.en.yaml
```

**Actual behavior**

```yaml
common:
  age_year: >-
    {count, plural,

      one {# year}
      other {# years}
    }
```

**Expected behavior**

```yaml
common:
  age_year: >-
    {count, plural,
      one {# year}
      other {# years}
    }
```

ritk7:
## Problem

`>-` (folded) scalars that contain a more-indented continuation line get a
spurious blank line inserted every time they're re-encoded, e.g.:

```
$ printf 'a: >-\n  foo(\n    bar\n  )\n' | yq .
a: >-
  foo(

    bar
  )
```

That extra blank line wasn't in the input. It reproduces #2409 (a `>-`
CI-query string gets a blank line inserted before an indented `select(...)`
continuation) and #2452 (a Symfony ICU plural message gains a blank line
after `sort_keys(..)`) — both are the same underlying bug, just triggered
via different expressions.

## Root cause

It's in the folded-scalar emitter in `go.yaml.in/yaml/v4` (yq's YAML
dependency), not in yq's own code. `writeFoldedScalar` decides whether a
line break needs to be doubled (so it survives folding when re-parsed) by
scanning forward from the break to see what follows it — but the scan
variable was initialized to `k := 0` instead of `k := i`, so it always
inspected the *start of the whole scalar* instead of the upcoming line. For
basically any folded scalar whose first character isn't blank, that made
the check always fire and insert an extra break.

This was already fixed upstream in
yaml/go-yaml@8edd87069cbecb411cb2f71f4c077a72b517e052 ("Fixes indexing bug
in emitter"), but no tagged release contains it yet — `v4.0.0-rc.6` is
still the latest tag, so Dependabot has had nothing to bump to.

## Fix

Bump `go.yaml.in/yaml/v4` to a pseudo-version pinned at that fix commit
(`go get go.yaml.in/yaml/v4@8edd87069cbecb411cb2f71f4c077a72b517e052`).
I checked the handful of commits between the current tag and this one and
they're small, unrelated fixes (a timestamp-unmarshal fix, a cyclical
LoadErrors reference fix, CI bumps) — nothing that touches encoding
behavior yq relies on.

Also added two regression cases to `yamlFormatScenarios` in
`pkg/yqlib/yaml_test.go` covering both issues' patterns. I verified they
fail against the old dependency version and pass against the new one.

## Testing

- `go test ./...` passes.
- Verified both original repros from #2409 and #2452 produce the expected
  output with this change.
- Confirmed the new tests fail without the bump (reverted it locally,
  reran) and pass with it.

Fixes #2409
Fixes #2452


ccoVeille:
That's not the way to do. Let's wait for a new version.

While this can be accepted by a core maintainer of yq because there is a need foe testing things in a dev branch, here it sounds a bold move

